PR #33, opened 2016-03-05
Implemented cbt run command

ChavXO:
Before "runClass" defaulted to returning "Main" so it didn't run the file. Still needs work but this a basic implementation.

### DIFF
--- a/stage2/Lib.scala
+++ b/stage2/Lib.scala
@@ -148,7 +148,7 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
       .filterNot(t => anyRefMembers contains taskName(t))
   }
 
-  class ReflectBuild(build: Build) extends ReflectObject(build){
+  class ReflectBuild(val build: Build) extends ReflectObject(build){
     def usage = {
       val baseTasks = lib.taskNames(ru.typeOf[Build])
       val thisTasks = lib.taskNames(subclassType) diff baseTasks
@@ -381,10 +381,12 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
     import scala.collection.JavaConversions._
     val watcher = WatchService.newWatchService
 
-    files.map{
-      file =>
-      if(file.isFile) dirname(file)
-      else file
+    val realFiles = files.map(realpath)
+
+    realFiles.map{
+      // WatchService can only watch folders
+      case file if file.isFile => dirname(file)
+      case file => file
     }.distinct.map{ file =>
       val watchableFile = new WatchableFile(file)
       val key = watchableFile.register(
@@ -398,13 +400,19 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
     scala.util.control.Breaks.breakable{
       while(true){
         logger.loop("Waiting for file changes...")
+        logger.loop("Waiting for file changes...2")
         Option(watcher.take).map{
           key =>
           val changedFiles = key
             .pollEvents
+            .toVector
             .filterNot(_.kind == StandardWatchEventKind.OVERFLOW)
             .map(_.context.toString)
+            // make sure we don't react on other files changed
+            // in the same folder like the files we care about
+            .filter{ name => realFiles.exists(name startsWith _.toString) }
             .map(new File(_))
+
           changedFiles.foreach( f => logger.loop( "Changed: " ++ f.toString ) )
           changedFiles.collect(action)
           key.reset
--- a/cbt
+++ b/cbt
@@ -212,17 +212,18 @@ stage1 () {
 			log "Checking if nailgun is up yet." $*
 			$NG cbt.NailgunLauncher cbt.CheckAlive $CP "$CWD" $* >> $nailgun_out 2>> $nailgun_err
 			alive=$?
-			if [ $alive -eq 131 ]; then
-				echo "Nailgun call failed. Try 'cbt kill' and check the error log cbt/nailgun_launcher/target/nailgun.stderr.log" 1>&2
-			elif [ $alive -eq 33 ]; then
+			if [ $alive -eq 131 ] || [ $alive -eq 33 ]; then
+			# the 33 is not working right now
+			#	echo "Nailgun call failed. Try 'cbt kill' and check the error log cbt/nailgun_launcher/target/nailgun.stderr.log" 1>&2
+			#elif [ $alive -eq 33 ]; then
 				break
 			else
 				log "Nope. Sleeping for 0.5 seconds" $*
 				if [ "$i" -gt "1" ]; then
 					echo "Waiting for nailgun to start... (For problems try -Dlog=nailgun or check logs in cbt/nailgun_launcher/target/*.log)" 1>&2
 				fi
-				sleep 0.5
 			fi
+			sleep 0.5
 		done
 		log "Running $mainClass via Nailgun." $*
 		$NG cbt.NailgunLauncher $mainClass $CP "$CWD" $*
--- a/stage2/Stage2.scala
+++ b/stage2/Stage2.scala
@@ -40,14 +40,16 @@ object Stage2{
 
         logger.loop("Looping change detection over:\n - "++allTriggerFiles.mkString("\n - "))
 
-        lib.watch(allTriggerFiles) {
+        lib.watch(allTriggerFiles){
           case file if triggerCbtFiles.exists(file.toString startsWith _.toString) =>
             logger.loop("Change is in CBT's own source code.")
             logger.loop("Restarting CBT.")
             scala.util.control.Breaks.break
 
           case file if triggerFiles.exists(file.toString startsWith _.toString) =>
-            new lib.ReflectBuild( lib.loadDynamic(context) ).callNullary(task)
+            val reflectBuild = new lib.ReflectBuild( lib.loadDynamic(context) )
+            logger.loop(s"Re-running $task for " ++ reflectBuild.build.projectDirectory.toString)
+            reflectBuild.callNullary(task)
         }
       } else {
         new lib.ReflectBuild(build).callNullary(task)


stage2/DefaultBuild.scala
```diff
@@ -214,7 +214,27 @@ class Build(val context: Context) extends Dependency with TriggerLoop{
     )
   }
 
-  def runClass: String = "Main"
+	def hasMainClass(file : File, classLoader : ClassLoader) : Boolean = {
```
farmdawgnation: This indentation doesn't look like what's been used throughout the rest of the code. :)

stage2/DefaultBuild.scala
```diff
@@ -214,7 +214,27 @@ class Build(val context: Context) extends Dependency with TriggerLoop{
...
+	  val className = getClassName(file)
+	  val inspectClass = classLoader.loadClass(className)
+	  inspectClass.getDeclaredMethods().map(_.getName).contains("main")
+	}
+	
+	
+	val files = compileTarget.listFiles
+	
+	val classFiles = files.filter(x =>x.toString.endsWith("class"))
```
farmdawgnation: You could use the `_` wildcard here to avoid having to use a single letter variable named `x`.

`files.filter(_.toString.endsWith("class"))`

Should we also ensure this `endsWith` checks to `.class` so that things like "Myawesomeclass" wouldn't match?

stage2/DefaultBuild.scala
```diff
@@ -214,7 +214,27 @@ class Build(val context: Context) extends Dependency with TriggerLoop{
...
+	
+	def getClassName(file : File) = file.getName.takeWhile(_ != '.')
+	
+	val hasMain = classFiles.filter(x => hasMainClass(x, classLoader))
```
farmdawgnation: Ditto here with replacing `x` with `_`.

`classFiles.filter(hasMainClass(_, classLoader))`

stage2/DefaultBuild.scala
```diff
@@ -214,7 +214,27 @@ class Build(val context: Context) extends Dependency with TriggerLoop{
...
+	
+  def runClass: String = try {
+	  getClassName(hasMain.head)
+	} catch {
+	  case e : NoSuchElementException => "Main"
```
farmdawgnation: What's the effect of returning `"Main"` here? Does it try to invoke a `Main` even if it doesn't exist?

farmdawgnation: Ah, this is essentially the existing behavior, isn't it?

ChavXO: Yes. Ideally though you would throw some kind of error. Not sure how dev envisions errors being propagated. Slightly reminiscent of SBT it would be nice, in the case that there are multiple files with a main method, if the tool allows the user to select which main to run. Working on that though.

farmdawgnation: Makes sense. :ok:

stage2/DefaultBuild.scala
```diff
@@ -214,7 +215,47 @@ class Build(val context: Context) extends Dependency with TriggerLoop{
...
+    ${mains.mkString("\t\n")}""")
+    //FIXME: doesn't stop to read the line!!!
+    //val choice = scala.io.StdIn.readLine() 
+    
+    files(0) // defaulting to first file in the meantime
+  }
+  
+  def getClassName(file : File) = file.getName.takeWhile(_ != '.')
+	
+  def runClass: String = {
```
cvogt: So the (undocumented) intention of this method was really just to determine the name of the class, that `def run` would run. See below.

I thought about if CBT should add a main class discovery mechanism and I am not sure. Maybe. In any case that logic should go into class Lib. This Build class is really just supposed to be a shallow layer on top of it and not contains major functionality as implemented here.

By the way, opening an interactive terminal seems to be non-standard in nailgun. I didn't investigate the details, but it would be good if someone gathers the knowledge how to do that, so we can also get rid of the publishSigned hack in the bash script that prevents running through nailgun in that case. Basically the problem is java.io.Console.getConsole is null.

cvogt: Yeah, I think it's a good idea actually. I had problem where I set the runClass to a non existing one and then cbt would just do nothing with the below logic.

cvogt: let's add it as a method to the library and figure out how interactive works.

cvogt:
this needs rebase on master and eventually before merging - squashing

stage2/DefaultBuild.scala
```diff
@@ -214,7 +215,47 @@ class Build(val context: Context) extends Dependency with TriggerLoop{
...
+    if (compileTarget == null) return "Main"
```
cvogt: why would it be null?

stage2/DefaultBuild.scala
```diff
@@ -214,7 +215,47 @@ class Build(val context: Context) extends Dependency with TriggerLoop{
...
+    val hasMain = getClassFiles(compileTarget)
+    
+	  val classFile = if (hasMain.length == 1) hasMain(0) else pickMain(hasMain) 
+	  try {
+      getClassName(classFile)
+    } catch {
+      case e : NoSuchElementException => "Main"
+    }
```
cvogt: this try catch seems unnecessary now, no?

stage2/DefaultBuild.scala
ChavXO: Trying to call cbt run before cbt compile. Ideally it should compile at that point.

cvogt: so `run` calls `classLoader`so this should work when called via run. Maybe just call compile here to make sure. Also compileTarget is a File object, it shouldn't be null.

stage2/DefaultBuild.scala
ChavXO: Yes. Spill-Over from previous code.

stage2/DefaultBuild.scala
```diff
@@ -214,7 +215,8 @@ class Build(val context: Context) extends Dependency with TriggerLoop{
     )
   }
 
-  def runClass: String = "Main"
+  def runClass: String = lib.getRunClass(compileTarget, classLoader)
```
cvogt: let's keep the names identical like everywhere else and get rid of the `get`

stage2/Lib.scala
```diff
@@ -65,6 +65,44 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
...
+      inspectClass.getDeclaredMethods().map(_.getName).contains("main")
+    }
+
+    def isInnerClass(file : File) : Boolean = file.toString.contains("$")
+    val files = target.listFiles
+
+    val classFiles = files.filter(file => file.toString.endsWith(".class") && !isInnerClass(file)).toList
+
+    classFiles.filter(hasMainClass(_))
+	}
```
cvogt: tab slipped in

stage2/Lib.scala
```diff
@@ -65,6 +65,44 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
...
+        
+    ${mains.mkString("\t\n")}""")
+
+    //FIXME: doesn't stop to read the line!!!
+    //val choice = scala.io.StdIn.readLine() 
+
+    files(0) // defaulting to first file in the meantime
+  }
+  
+  def getClassName(file : File) = file.getName.takeWhile(_ != '.')
```
cvogt: how about making all of the above inner functions of getRunClass, to underline that that aren't used by anything else and keep the exported api small? If needed we can move them out again later.

stage2/Lib.scala
```diff
@@ -65,6 +65,44 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
...
+    val classFiles = files.filter(file => file.toString.endsWith(".class") && !isInnerClass(file)).toList
+
+    classFiles.filter(hasMainClass(_))
+	}
+  
+  def pickMain(files : List[File]) : File = {
+    val mains = files.map(_.getName).zipWithIndex map {case (fileName, index) => s"""[${index + 1}] ${fileName}\n"""}
+    println(s"""Multiple main classes detected. Select one to run:
+        
+    ${mains.mkString("\t\n")}""")
```
cvogt: does this indent all but the first one with a tab? I think we don't need to indent them. Just scratch the \t :)

cvogt:
How about this for a design:

cvogt:
```
// class BasicBuild
def run: Unit = lib.run( mainClasses, Seq(), classLoader )  )
def mainClass: Option[String] = mainClasses.headOption
def mainClasses: Seq[String] = lib.mainClasses(compileTarget, classLoader)
```

```

**Comment by ChavXO** [comment]
@cvogt So  the default behaviour would be to run all main classes found in the target directory?

**Comment by cvogt** [comment]
no, only if there is exactly one. Actually the above runs simple one of them, but none in case of ambiguity is better I suppose

**Comment by cvogt** [comment]
fixing the code above

**Comment by cvogt** [comment]
.foreach is the foreach on Option there, not on Seq

**Comment by ChavXO** [comment]
@cvogt Done.

File: stage1/Stage1.scala [hunk 12]
```diff
@@ -5,6 +5,7 @@ import java.time.LocalTime.now
 
 import scala.collection.immutable.Seq
 import scala.collection.JavaConverters._
+import scala.language.postfixOps
```
cvogt: did you add this by accident?

stage2/BasicBuild.scala
```diff
@@ -144,8 +147,16 @@ class Build(val context: Context) extends Dependency with TriggerLoop{
     )
   }
 
-  def runClass: String = "Main"
-  def run: ExitCode = lib.runMainIfFound( runClass, context.args, classLoader ) 
+  def run: ExitCode = lib.run( runClass, classLoader )
+  def runClass: Option[String] = {
+	  if (mainClasses.size == 1)
```
cvogt: There is a tab in here again. I suggest you find some way to let your editor know.

stage2/BasicBuild.scala
```diff
@@ -105,6 +106,8 @@ class Build(val context: Context) extends Dependency with TriggerLoop{
   //def cacheJar = false
   override def dependencyClasspath : ClassPath = ClassPath(localJars) ++ super.dependencyClasspath
   override def dependencyJars      : Seq[File] = localJars ++ super.dependencyJars
+  
+  def clean : ExitCode = lib.clean(target)
```
cvogt: let's do the clean stuff in a separate PR. git cherry pick will help migrate the commit to another branch.

stage2/Lib.scala~
```diff
@@ -0,0 +1,474 @@
...
+            .map(_.context.toString)
+            .map(new File(_))
+          changedFiles.foreach( f => logger.loop("Changed: "+f) )
+          changedFiles.collect(action)
+          key.reset
+        }
+      }
+    }
+  }
+}
```
cvogt: This file slipped in, but shouldn't have.

stage2/Lib.scala
```diff
@@ -66,6 +66,52 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
...
+      println("[info] Failed")
+      return ExitCode.Failure
+    }
+  }
+
+  
+  def run( mainClass: Option[String], classLoader: ClassLoader ) : ExitCode = mainClass match {
+	  case Some(className) =>  runMain( className, Seq(), classLoader )
+	  case _               => ExitCode.Failure
+	}
```
cvogt: tabs

cvogt: also common pattern here is to use getOrElse

```scala
mainClass.map( runMain( _, Seq(), classLoader ) ).getOrElse( ExitCode.Failure )
```

stage2/Lib.scala
```diff
@@ -66,6 +66,52 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
...
+  
+  def mainClasses(compileTarget : File, classLoader : ClassLoader): Seq[String] = {
+    def className(file : File) = file.getName.takeWhile(_ != '.')
+    
+    def classesWithMain(target : File, classLoader : ClassLoader) : List[File]  = { 
+      def hasMainClass(file : File) : Boolean = {
+        val name = className(file)
+    
+        val inspectClass = classLoader.loadClass(name)
+        inspectClass.getDeclaredMethods().map(_.getName).contains("main")
```
cvogt: this should either check if the argument list has exactly one element of type Array[String] or add a comment here that that's TODO.

stage2/BasicBuild.scala
cvogt: the thing with headOption is, you don't need to check for the size.

stage2/Lib.scala
```diff
@@ -66,6 +66,52 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
...
+      }
+
+      def isInnerClass(file : File) : Boolean = file.toString.contains("$")
+      val files = target.listFiles
+
+      val classFiles = files.filter(file => file.toString.endsWith(".class") && !isInnerClass(file)).toList
+
+      classFiles.filter(hasMainClass(_))
```
cvogt: why not return name here? If hasMainClass would be called mainClass and returns an Option[String], you could use flatMap here instead of filter and save yourself from calling .map(className(_)) again below. Also the name hasMain below is not accurate and would not be needed anymore then.

stage2/BasicBuild.scala
cvogt: would be good if we could also add a test for this. could be one of the sample builds in there. compile it, check target exists, then clean, check it doesn't exist.

cvogt:
would be good to get a test for this as well. Check test/ and just add builds with one main method, multiple main methods and check the results are correct. Thx Michael :).

ChavXO:
@cvgot Moved. Working on tests.

Issue #25, opened 2016-03-05
clean up source code


cvogt:
there has been some superficial dirtiness been introduced in the last two days enhancements under time pressure. Let's clean that up.

cvogt:
closed in favor of https://github.com/cvogt/cbt/issues/25

PR #36, opened 2016-03-05
Some style cleanups in Stage1.

farmdawgnation:
Some suggestions for #25. Probably some more things I'll fine when I get another chance to sit down and tinker with it, but I wanted to open early and provide some opportunity for feedback before I've invested a day in it. :grin:

### DIFF
--- a/stage1/Stage1.scala
+++ b/stage1/Stage1.scala
@@ -1,70 +1,81 @@
 package cbt
+
 import java.io._
-import paths._
+import java.time.LocalTime.now
+
 import scala.collection.immutable.Seq
+import scala.collection.JavaConverters._
+
+import paths._
 
 object CheckAlive{
   def main(args: Array[String]): Unit = {
     System.exit(33)
   }
 }
+
+class Init(args: Array[String]) {
+  /**
+   * Raw parameters including their `-D` flag.
+  **/
+  val propsRaw: Seq[String] = args.toVector.filter(_.startsWith("-D"))
+
+  /**
+   * All arguments that weren't `-D` property declarations.
+  **/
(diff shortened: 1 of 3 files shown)


stage1/Stage1.scala
```diff
@@ -1,70 +1,81 @@
...
+
+import paths._
 
 object CheckAlive{
   def main(args: Array[String]): Unit = {
     System.exit(33)
   }
 }
+
+class Init(args: Array[String]) {
```
cvogt: We should probably start making things private[cbt] in order to reduce the exposed interface

stage1/Stage1.scala
```diff
@@ -1,70 +1,81 @@
...
+  /**
+   * Raw parameters including their `-D` flag.
+  **/
```
cvogt: If it's a one liner, I generally prefer the more compact
```
/** Raw parameters including their `-D` flag. */
```

cvogt:
This is great. LGTM

stage1/Stage1.scala
farmdawgnation: Good idea.

stage1/Stage1.scala
farmdawgnation: Ah, sure. That makes sense.

cvogt: actually, I take that back. let's not get into whitespace discussions. once there is scalariform, that's all unified anyways.

stage1/Stage1.scala
cvogt: dropped it again in order to access it from the tests. maybe those should go into the same package. they test via processes not jvm anyways

Issue #35, opened 2016-03-05
Different indentation


farmdawgnation:
There are a few different indentation strategies used throughout the source. Some things are using tabs, others are using spaced. @cvogt Which is intended? :)

cvogt:
@farmdawgnation there are tabs in there? Not indented. In support for cleaning up indentation in general. Let's coordinate to avoid merge conflicts. Also in favour of Scalariform, if we find a nice way to integrate it, maybe a cbt build ;)

cvogt:
these are my preferred settings
        .setPreference(AlignParameters, true)
        .setPreference(AlignArguments, true)
        .setPreference(AlignSingleLineCaseStatements, true)
        .setPreference(MultilineScaladocCommentsStartOnFirstLine, true)
        .setPreference(SpaceInsideParentheses, true)
        .setPreference(SpacesWithinPatternBinders, true)
        .setPreference(SpacesAroundMultiImports, true)
        .setPreference(DoubleIndentClassDeclaration, false)

farmdawgnation:
I'm down for scalariform. Though that won't help tabs in shell scripts and java files. ;)

cvogt:
so the tabs in Java have been replaced by spaces. I am closing this.

Issue #32, opened 2016-03-05
Cleanup file structure


cvogt:
mostly put things into files named after the classes to make stuff easier to find

PR #41, opened 2016-03-06
bootstrap/nailgun minor cleanup

tpolecat:
A few minor changes.

- whitespace, etc., for readability
- move `Dependency` into its own compilation unit
- propagate checked exceptions rather than wrapping

For Java code I prefer to propagate checked exceptions in the method types rather than wrapping them in `RuntimeException` or simply marking everything as `throws Exception`. As code evolves this helps prevent inadvertent propagation of exceptions that can actually be handled; the programmer gets no indication due to the outer catch-all.

### DIFF
--- a/bootstrap_scala/BootstrapScala.java
+++ b/bootstrap_scala/BootstrapScala.java
@@ -1,105 +1,79 @@
-import java.io.*;
-import java.lang.reflect.*;
-import java.math.*;
-import java.net.*;
-import java.nio.*;
-import java.nio.file.*;
-import java.security.*;
-import java.util.*;
-import java.util.stream.*;
-import javax.tools.*;
+import java.io.File;
+import java.io.InputStream;
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
 import javax.xml.bind.annotation.adapters.HexBinaryAdapter;
 
-/*
-This file class allows bootstrapping out of Java into Scala.
-It downloads the Scala jars for the version number given as the first
-argument into the directory given as the second argument and returns
-a classpath String.
-*/
-public class BootstrapScala{
-	public final static Dependency[] dependencies(String target, String scalaVersion){
-		return new Dependency[]{
-			Dependency.scala(target, scalaVersion, "library", "f75e7acabd57b213d6f61483240286c07213ec0e"),
-			Dependency.scala(target, scalaVersion, "compiler","1454c21d39a4d991006a2a47c164f675ea1dafaf"),
-			Dependency.scala(target, scalaVersion, "reflect", "bf1649c9d33da945dea502180855b56caf06288c"),
-			new Dependency(target, "modules/scala-xml_2.11/1.0.5", "scala-xml_2.11-1.0.5", "77ac9be4033768cf03cc04fbd1fc5e5711de2459")
-		};
-	}
+/**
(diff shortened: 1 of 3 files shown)


bootstrap_scala/BootstrapScala.java
```diff
@@ -1,105 +1,79 @@
...
 
-class Dependency{
-	URL url;
-	Path path;
-	String hash;
-	public Dependency(String target, String folder, String file, String hash){
-		path = Paths.get(target+file+".jar");
-		try { url = new URL("https://repo1.maven.org/maven2/org/scala-lang/"+folder+"/"+file+".jar"); }
-		catch (final MalformedURLException e) { throw new RuntimeException(e); }
-		hash = hash;
```
tpolecat: This was a bug; it was simply reassigning the argument to itself and the `hash` member remained null. Marking the fields as `final` exposed this.

tpolecat:
re: #32 #35

bootstrap_scala/Dependency.java
```diff
@@ -0,0 +1,29 @@
...
+  final String hash;
+
+  public Dependency(String target, String folder, String file, String hash) throws MalformedURLException {
+    this.path = Paths.get(target + file + ".jar");
+    this.url  = new URL("https://repo1.maven.org/maven2/org/scala-lang/" + folder + "/" + file + ".jar");
+    this.hash = hash;
+  }
+
+  // scala-lang dependency
+  public static Dependency scala(String target, String scalaVersion, String scalaModule, String hash)
```
farmdawgnation: I'm a bit bothered - maybe for no good reason - that instantiating a Java dependency requires `new Dependency` and instantiating a Scala dependency is `Dependency.scala`. Another option would be to protect the constructor and add `Dependency.java` so our interface is symmetrical.

That has little to no practical effect, I think, so if you hate it then feel free to ignore me. :)

farmdawgnation:
Didn't check it out and test it, but it looks good on scan. :)

bootstrap_scala/Dependency.java
cvogt: In case you missed that. bootstrap_scala is tiny a self-contained scala jar downloader, that integrates with the rest of cbt via processes, not via the jvm. It never ends up on the same classpath as CBT itself. So I don't think the internal API of bootstrap_scala matters much.

(We tried but failed at recording the architecture session at NEScala. I'll try to do a screen cast one of these days.)

bootstrap_scala/BootstrapScala.java
```diff
@@ -1,105 +1,79 @@
...
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
```
cvogt: I am generally a fan of wild-card imports, because I don't see the value of spending time micro-managing imports and I don't use an IDE that does it for me. I very rarely had bugs because of that. When I did, it was with the areas where Scala's type-safetly is lacking, in particular Any2StringAdd. I think I'd rather try to enforce using ++ everywhere than micromanaging imports, especially when they are all JDK ones. WDYT?

tpolecat: I use wildcard imports when I'm trying to figure out what I'm doing, but I usually turn them into explicit imports as part of general tidying up. To me it's just another source of confusion that I can minimize with a minute or two of work. I don't use an IDE either, so the imports help me figure out where things are coming from.

But you're driving the bus so I can revert this to wildcards if you like.

cvogt: na, let's not waste time on going back. The only way is forward ;). And the Java code should be more stable than the rest of the project. LGTM

cvogt:
+1 for propagating checked exceptions

PR #44, opened 2016-03-06
integrate circleci

cvogt:


farmdawgnation:
Haha. I started work on doing this with Travis before my flight. Does circle support OSX testing too?

cvogt:
I don't know. I have no strong opinion on which CI we use. I just knew how to make circle happen. Very open to adding others.

farmdawgnation:
Got it. I'm not a huge fan of Circle but if it's what you and others know how to use it's fine. If it causes problems switching from one to the other is always trivial.   ¯\_(ツ)_/¯

cvogt:
does travis support osx?

farmdawgnation:
It does. It looks like Circle does too, but their tooling looks to be explicitly around iOS testing from my first few google searches. I admittedly didn't do a ton of research because I'm in an airport. :)

cvogt:
Anything works. I just wanted to get SOMETHING up quickly. Still working on making it happen.

cvogt:
Not quire circle support yet, but working towards it

Issue #54, opened 2016-03-07
ArrayIndexOutOfBoundsException in NailgunLauncher


tobias-johansson:
Clean bootstrap on 5488e70 (problem not present on c2f53c5)
(Ubuntu, javac 1.8.0_66-internal, OpenJDK Runtime Environment (build 1.8.0_66-internal-b17))

```bash
> cbt run
(Note: nailgun not found. It makes CBT faster! Try 'brew install nailgun'.)
Recompiling CBT BootstrapScala.java
downloading https://repo1.maven.org/maven2/org/scala-lang/scala-library/2.11.7/scala-library-2.11.7.jar
to /home/tobias/Projects/cbt/bootstrap_scala/cache/2.11.7/scala-library-2.11.7.jar
downloading https://repo1.maven.org/maven2/org/scala-lang/scala-compiler/2.11.7/scala-compiler-2.11.7.jar
to /home/tobias/Projects/cbt/bootstrap_scala/cache/2.11.7/scala-compiler-2.11.7.jar
downloading https://repo1.maven.org/maven2/org/scala-lang/scala-reflect/2.11.7/scala-reflect-2.11.7.jar
to /home/tobias/Projects/cbt/bootstrap_scala/cache/2.11.7/scala-reflect-2.11.7.jar
downloading https://repo1.maven.org/maven2/org/scala-lang/modules/scala-xml_2.11/1.0.5/scala-xml_2.11-1.0.5.jar
to /home/tobias/Projects/cbt/bootstrap_scala/cache/2.11.7/scala-xml_2.11-1.0.5.jar
Recompiling NailgunLauncher. Detected source changes.
Stopping nailgun
Restarting nailgun
Recompiling Stage1. Detected source changes.
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 3
	at cbt.NailgunLauncher.main(NailgunLauncher.java:42)
```

cvogt:
fixed

PR #55, opened 2016-03-07
Circleci

cvogt:


cvogt:
fixes #54. Seems like there is still something not working correctly which prevents from connecting to nailgun in certain circumstances.

PR #68, opened 2016-03-08
Clean compiled files

ChavXO:


ChavXO:
@cvogt all that's left is adding the tests.

cvogt:
@ChavXO and removing the run stuff. and squashing

PR #73, opened 2016-03-09
Implemented run command allowing for package discovery.

mchav:
@cvogt Fresh implementation of run. With new text editor after getting through some git tutorials. :)


stage2/BasicBuild.scala
```diff
@@ -144,8 +144,10 @@ class Build(val context: Context) extends Dependency with TriggerLoop{
     )
   }
 
-  def runClass: String = "Main"
-  def run: ExitCode = lib.runMainIfFound( runClass, context.args, classLoader ) 
+  def run: ExitCode = lib.run( runClass, classLoader )
+  def runClass: Option[String] = mainClasses.headOption
```
cvogt: `def mainClass` please :). (change and then use `git commit -a --amend`. After that `git push --force`)

cvogt: actually nevermind. This is good. I'll do minor changes later.

ChavXO:
@cvogt +1

PR #74, opened 2016-03-10
Refactor cbt build script...

hntd187:
Working on a bit of this along the way, got a chance to rewrite a lot of this to be what I think is smarter and a bit more performant, but a test is still failing, would someone be so kind as to try this out and let me know how we can fix it?

### DIFF
--- a/DEVELOPER_GUIDE.txt
+++ b/DEVELOPER_GUIDE.txt
@@ -1,9 +1,8 @@
 Welcome developer.
 
-CBT has a very easy code base that you can fully master in under an hour.
-Don't shy away to submit PRs :).
-
-CBT launches in stages in order to bootstrap from source out of Java into Scala and finally CBT.
+CBT has a very easy code base that you can fully master it in an afternoon.
+Don't shy away from submiting PRs :). And because CBT bootstraps from source
+you already have the code there.
 
 The ./cbt bash script starts the process.
 
@@ -19,3 +18,9 @@ stage1/	            CBT's code that only relies only on Scala/Java built-ins. Co
 stage2/             CBT's code that requires additional libs, e.g. barbary watchservice.
 test/               Unit tests that can serve as example builds
 sonatype.login      Sonatype credentials for deployment. Not in git obviously.
+
+CBT follows an optimistic merging approach. (See http://hintjens.com/blog:106).
+We strongly suggest well polished PRs, but don't want to stall improvements
+by discussions about minor flaws. As long as the PR does not break anything and
+improves the product, we should merge it, polishing remaining things afterwards.
+
--- a/README.md
+++ b/README.md
@@ -118,12 +118,9 @@ Missing features in comparison with SBT
 ---------------------------------------
 
 Not implemented yet, but rather easily possible without API changes or
-major refactors is concurrently building / downloading dependencies and
+major refactors is concurrently building dependencies and
 running tests. Right now it is sequential.
-
-CBT allows tasks to lazily depend on other tasks.
-SBT currently does not, because it uses an Applicative, not a Monad, so
-task dependencies are eager.
+Maven downloads already happen in parallel as well as some tasks like packaging.
 
 Another edge case that may need a solution is dynamically overwriting
 tasks. SBT allows that. Classes and traits are static. The only use
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 Join the chat at https://gitter.im/cvogt/cbt
 
 Welcome to Chris' Build Tool (CBT) for Scala
+============================================
 
 Fun, fast, intuitive, composable and statically checked builds written in Scala.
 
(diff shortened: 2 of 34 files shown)


hntd187:
Okay I massively screwed this up, I thought my fork was synced with master, but apparently it is not. I have to go back and redo a lot of this.

hntd187:
Okay @cvogt I'm gonna close this request and redo this. I'm gonna implement a lot of the new build stuff into what I have and resubmit this. If you can comment at least on structure and style, I think it's much cleaner and has some better ideas than the way it's written, but I'm open to whatever. I also suggest we either choose if we are gonna use bashisms or try and stick to non-bashisms. I mean stuff like [[ vs [ for test and back ticks vs $() you'll see I used some of that stuff in my script, I think it's cleaner and I believe it performs a bit better, but I'm open to discussion about it.

cvogt:
@hntd187 oh, sorry to head we conflicted here :-\. Love the cleanup though. One experiment we should do is: what is the performance if we write the whole thing in Java and let the shell script only conditionally compile it and run it.

Sounds kindof counter-intuitive, because we are starting another JVM just to talk to an already running one that uses Nailgun. Doesn't this break the whole idea of Nailgun? In my initial experiments it didn't. Vanilla JVM startup time without classloading anything except for one single script was about 0.1 seconds, which is lower than the time the bash script currently runs until it hands over to nailgun. ClassLoading seems to be the expensive part, but that launcher JVM wouldn't have to do much. So we might actually get safer, cleaner code and a speed up writing it in Java instead of bash. But that would need to be tested.

Alternatively we should be able to gain some clarity by moving as much as we can into Java, mostly compiling stage1 and doing that from NailgunLauncher instead of shell. And on the shell level use functions for code re-use. I also like line-breaks for things with many parameters, e.g. the call to the compiler.

WDYT?

hntd187:
@cvogt I think while we are distributing by source it doesn't make a lot of sense to do something like that. Especially when bash overall is pretty low overhead, so I think what we do now with the bootstrapper and all that seems to be a bit overkill, but again it's just my opinion. Basically your bootstrapper is the beginnings of what I'd imagine a wrapper like gradle would be like, but that's one of those things that would be more down the line when we start doing binaries and distribution and such. I agree I think nailgun is where everything should be and either you use it for everything or don't use it at all, I think having 2 JVMs going just for running builds seems to be unnecessarily redundant for no obvious benefit. 

I've been rewriting some parts I think are too heavy at this point for a source based distribution, but again I'll defer the end choice to you. But you can look at what I'm working on in my fork.

https://github.com/hntd187/cbt/blob/master/cbt

Some things,
1. I moved all the POSIX esq stuff to bash, but I need some people to test it to make sure it works, but for the most part I'm not worried about it because I feel like cbt would be a tool used on 10 year old distros of BSD.
2. I rewrote the scala_bootstrapper in bash. It's quick, easy and I think it looks a little better and is faster, but you can make the call if you think overall it's a good idea.
3. I added the quiet word to supress some of the script output, but I'm still testing it so it's not really finished, but still looking for opinions.

Thoughts?

cvogt:
@hntd187 How is this related to source distribution or not? (This may be a permanent distribution form by the way, but paired with some versioning concept)

My personal feeling that bash is weird and fragile and only interactive performance could potentially justify it's use.

Wrapper or not, I am mostly concerned about performance that occurs on every startup. If everything has been compiled previously, cbt right now takes about 0.6 seconds on the most trivial tasks on my machine, which feels relatively interactive. My second concern is correctness and maintainability of the launcher, which Java would improve. The only question is if there would be a performance tax and what would it be.

I am not worried about the cost for compiling once if it is cached afterwards. Correctness and maintainability trumps that one time cost significantly in my eyes. So I am skeptical about re-writing bootstrap_scala in bash. (The checksum checking was in there by the way because I had encountered corrupted downloads. I doubt we could go without it.)

I would also like to keep dependencies as few as possible. curl is one more, which is probably available on most unixes, but wouldn't require it if there it a non inferior way without it. That's why I wrote the downloader in Java.

hntd187:
@cvogt Well, I'll agree that bash is totally weird, but it does seem rather overkill to compile a java file and start a JVM every time just to do file existence checking. Was corrupt downloads a reoccurring problem for you? I honestly can't remember the last time I downloaded a file via curl or anything that had a problem with corrupt downloads, but that's just me. Bash is pretty good and consistent for the kinds of things system things we're doing here and java doesn't really add any extra safety just because it does does a checksum check, that could just as easily be in bash. 

To compare, sbt if you don't download and install a system wide distro has a 500 line shell script you can use from sbt-extras, which I've noticed is used in a lot of places because then you can just distribute that script and that script can do the minor things required to get the source code to build. 
https://github.com/paulp/sbt-extras/blob/master/sbt
vs
https://github.com/sbt/launcher

Also while I use curl, it's very easy to have a backup solution via wget, which if a unix based system that doesn't come with either wget or curl, then I dunno it must be a system with an insanely minimal installation or from the stone age.
https://curl.haxx.se/docs/osdistribs.html

But let's back up for a second. I think I'm starting to mesh 2 different requirements from this so let's firm up what we're trying to do here. If we are going to stick to source code distribution then, I think having this in bash will reduce the initial entry costs for distribution and will be something reproducible on most every unix system, but if we are going to eventually move beyond source code then we should be moving this more into something type checked like java or scala and use bash as a simple wrapper to obtain and run that code. It's what gradle and sbt use for their binary distros if you don't do a system wide install. 

I mean honestly I really think either solution is perfectly fine and I don't think either solution for us is necessarily a massive performance hit, so I am suggesting we either move everything into java/scala and the script compiles and runs that, or we do it mostly in bash and worry about moving it over when we think we may have outgrown it. Hopefully that makes sense.

cvogt:
How about we go with bash script refactoring for now and see when we outlive it :)?

cvogt:
(please keep bootstrap_scala as is though, I am happy to have it in Java)

cvogt:
@hntd187 did you already put any more work into this? I need to make some changes to the launcher

hntd187:
@cvogt no I didn't I fell asleep so I didn't get a chance to do anything.

Also I wanted to suggest, how about we throw the boostrap scala in, but let's keep the bash stuff in as well so if the bootstrap scala fails for whatever reason we can still bootstrap it. It's only like 20 lines or so of bash, so maybe it's an okay fallback?

cvogt:
@hntd187 Not in favor of duplication. Let's keep CBT small and focused. Also, let's pause this refactoring for now, so we don't conflict, because I have to do some changes to the script for other reasons in the next few days. We can get back to it at a later stage.

hntd187:
@cvogt kk :+1: I just figured the redundancy might not hurt, but you're right, best to keep these things simple.

PR #79, opened 2016-03-12
A draft implementation that runs builds concurrently

cvogt:
This is what I came up with for concurrent builds in CBT. I realized last minute that we probably don't need the KeyLockedLazyCache, if we just use member-level caching again. I commented it out for now. Opinions?

### DIFF
--- a/stage1/KeyLockedLazyCache.scala
+++ b/stage1/KeyLockedLazyCache.scala
@@ -1,0 +1,33 @@
+/*
+package cbt
+import java.util.concurrent.ConcurrentHashMap
+import scala.concurrent.Future
+
+/**
+A cache that lazily computes values if needed during lookup.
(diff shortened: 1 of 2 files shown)


stage1/KeyLockedLazyCache.scala
```diff
@@ -0,0 +1,33 @@
...
+    // not block the whole cache, but just that cache entry
+    key.synchronized{
+      if( ! (builds containsKey keyObject) ){
+        builds.put( keyObject, value )
+      }
+      builds get keyObject      
+    }
+  }
+}
+*/
```
cvogt: so github's scala highlighter is apparently broken on nested comments

cvogt:
merging this now to avoid merge conflicts. it shouldn't interfere with anything. We can review and fix as we go along

Issue #80, opened 2016-03-12
trap exit codes of user code


cvogt:
if user code does System.exit we may want to catch that in CBT rather than shutting down the jvm. There currently is `trapExitCode` in CBT which turns System.exit into an exception and catches it. But it is disabled because it lead to weird side-effects where the exception throwing security manager stuck around and threw exceptions beyond the trapping scope.

SBT seems to do something similar with it's `sbt.TrapExitSecurityException`. Maybe we can learn from them how to do it right.

PR #82, opened 2016-03-12
Various exit code related fixes.

### LINKED ISSUE #76: properly detect if zinc succeeded or not.
Currently it doesn't seem to communicate this via it's exist code.

That means we can't cache results, because we don't know if they succeeded. That leads to more calls to zinc than needed.

We also have to hope it succeeded when executing run and if it didn't we get exceptions.

A solution could be not using zincs runner. Probably the best way to do this is not calling zinc in stage1, but simply downloading it (using scalac to compile stage 2 instead). Then stage 2 can use zinc's typesafe api to configure it in a way that works as we need it.

farmdawgnation:
I believe this closes #76 and #80 in one go. This should also fix the issues with our CI not detecting that our tests have failed because those error codes should be properly bubbling too. There are also a handful of whitespace fixes because my editor does that automatically.

I also did a bit of cleanup of `zinc(...)`

### DIFF
--- a/stage1/Stage1Lib.scala
+++ b/stage1/Stage1Lib.scala
@@ -131,7 +131,7 @@ class Stage1Lib( val logger: Logger ) extends BaseLib{
 
     // only run zinc if files changed, for performance reasons
     // FIXME: this is broken, need invalidate on changes in dependencies as well
-    if( true || needsRecompile ){
+    if( needsRecompile ){
       val zinc = MavenDependency("com.typesafe.zinc","zinc", zincVersion)(logger)
       val zincDeps = zinc.transitiveDependencies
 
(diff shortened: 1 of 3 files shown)


cvogt:
great stuff :).

farmdawgnation:
Uh weird. The cbt direct command failed for circle but not for me locally.

cvogt:
@farmdawgnation one thought: the test may be wrong ;)

farmdawgnation:
Looks like it worked on #79.... Judging by the output the wrong params got passed to scalac. Can you try rebuilding? I don't think I can trigger a rebuild here since I'm not a collaborator.

cvogt:
Closed in favor of https://github.com/cvogt/cbt/pull/82

PR #96, opened 2016-04-02
Scaffold to start integrating tut.

cvogt:


cvogt:
@tpolecat it seems like there is still a bug with custom build time dependencies. cbt's build file compiles fine, but crashes with a class not found exception for TutMain when run. I'll check it out next.

cvogt:
force pushed Tut plugin

PR #100, opened 2016-04-04
Windows compatibility

mchav:
Added a batch file translation of `cbt`.


michaelyaakoby:
Hi, looking at the cbt.bat I see it is using for nc.exe, ng.exe and ng-server.exe (which I assume are netcat, nailgun-client and nailgun-server).
Will it be possible for you to share the binaries you use (i already had nc.exe, but couldn't find nailgun binaries).
BTW, maybe we can use chocolatey to distribute those binaries.
Thanks

cvogt:
closed in favor of #136 which only merges the .bat script

@mchav replacing the / with a generic path separator is a good idea, but we should use `java.io.File.separator` instead and there are many more places.

@michaelyaakoby the binaries are in the zinc repository linked here: https://github.com/cvogt/cbt/issues/122

PR #101, opened 2016-04-04
Clean command

mchav:
Clean command with tests.


test/test.scala
```diff
@@ -124,10 +134,13 @@ object Main{
 
     usage("nothing")
     compile("nothing")
+    clean("nothing")
```
cvogt: I am worried about this. If we mess up the compileTarget during development, this may end up deleting the entire hard drive.

stage2/Lib.scala
```diff
@@ -53,6 +53,31 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
     }
   }
 
+
+  /* recursively deletes folders*/
+  def deleteRecursive(file: File) : Boolean = {
+    if (file.isDirectory) {
+      file.listFiles().map(deleteRecursive(_))
```
cvogt: this leaves the empty directories in existence, right? Maybe we should delete those, too?

stage2/Lib.scala
```diff
@@ -53,6 +53,31 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
...
+    else {
+      deleteIfExists(file.toPath)
+    }
+  }
+  
+  def clean (compileTarget : File) : ExitCode = {
+    logger.lib(s"""Cleaning ${compileTarget}""")
+    if (!compileTarget.exists) return ExitCode.Success
+    
+    if (deleteRecursive(compileTarget)) {
```
cvogt: I wonder if at least an interactive user prompt asking if `compileTarget` should be deleted may make this less scary? We should probably also add checks like compileTarget != "" and maybe others.

cvogt: or maybe even better, just return a new line separated list of paths to delete. then users can review and optionally pass to | xargs rm -rf

PR #103, opened 2016-04-07
cross building and automatic snaphot ids

cvogt:
Adds support for cross building for multiple scala versions
And for automatically appending -SNAPSHOT to the id, when publishing a snapshot.

### DIFF
--- a/stage2/Lib.scala
+++ b/stage2/Lib.scala
@@ -24,6 +24,11 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
   val buildClassName = "Build"
   val buildBuildClassName = "BuildBuild"
 
+  def copy(cls: Class[_], context: Context) = 
+    cls
+    .getConstructor(classOf[Context])
+    .newInstance(context)
+
   /** Loads Build for given Context */
   def loadDynamic(context: Context, default: Context => Build = new Build(_)): Build = {
     context.logger.composition( context.logger.showInvocation("Build.loadDynamic",context) )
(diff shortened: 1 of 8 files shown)


cvogt:
@tpolecat @farmdawgnation could use a review on this one. It add support for cross compiling/packaging via `cbt cross compile` and `cbt cross package`. WDYT?

farmdawgnation:
I can take a look at this later tonight if nobody gets to it before me.

stage2/Lib.scala
```diff
@@ -53,16 +58,16 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
     }
   }
 
-  def srcJar(sourceFiles: Seq[File], artifactId: String, version: String, jarTarget: File): Option[File] = {
+  def srcJar(sourceFiles: Seq[File], artifactId: String, scalaMajorVersion: String, version: String, jarTarget: File): Option[File] = {
     lib.jarFile(
-      jarTarget ++ ("/"++artifactId++"-"++version++"-sources.jar"),
+      jarTarget ++ ("/"++artifactId++"_"++scalaMajorVersion++"-"++version++"-sources.jar"),
```
farmdawgnation: Are we concerned about compatibility with the pre 2.10 series when the full version was required in build artifact names?

cvogt: I thought about this. It is still now required now for releases that require binary compatibility like ammonite repl. So in general yes, but maybe not right away? seems simple enough to add later

stage2/Stage2.scala
```diff
@@ -15,8 +15,9 @@ object Stage2 extends Stage2Base{
     logger.stage2(s"Stage2 start")
     val loop = args.args.lift(0) == Some("loop")
     val direct = args.args.lift(0) == Some("direct")
+    val cross = args.args.lift(0) == Some("cross")
```
farmdawgnation: So, just to check my understanding... to do a cross compile I would need to do...

```
cbt cross compile
```

But that works with _any_ task. So I could do...

```
cbt cross <anything>
```

Right?

stage2/Stage2.scala
```diff
@@ -27,6 +28,18 @@ object Stage2 extends Stage2Base{
...
 
+    def call(build: Build) = {
+      if(cross){
+        build.crossScalaVersions.foreach{
+          v => new lib.ReflectBuild(
+            build.copy(context.copy(scalaVersion = Some(v)))
+          ).callNullary(task)
+        }
+      } else {
+        new lib.ReflectBuild(build).callNullary(task)
```
farmdawgnation: What is the scala version in this case? `defaultVersion`? Is there a possibility it could be unset or anything wacky like that?

stage2/Lib.scala
farmdawgnation: Got it. Hmmm. We should probably make a note of that somewhere visible if we haven't already. Just so we don't make someone accidentally waste an afternoon tracking it down.

farmdawgnation:
Code seems straightforward enough. Admittedly, I haven't been following along too closely over the past few weeks - but it feels pretty good that I can still navigate the codebase in spite of that. Probably a good sign for simplicity. :)

stage2/Stage2.scala
cvogt: yes

stage2/Stage2.scala
cvogt: it defaults to cbt's own scala version. Users have to override it. If they put something weird or not a full version, e.g. 2.10 instead of 2.10.6, weird thing will likely happen. Let's deal with these as they come and put a focus on providing good error messages or at least and FAQ that covers these cases in the docs

farmdawgnation: :+1:

PR #104, opened 2016-04-07
tut proof of concept

tpolecat:
Need to make a few tweaks to **tut** so we can call it in-process more cleanly, but this does work. I'm not sure what we want to have on the classpath here or how to get it; I'm just sending the `dependencyClasspath` for now.

Note that in real life we will need to run in a new classloader with the target version of scala and so forth. Baby steps. It's much much easier than calling from sbt.

### DIFF
--- a/build/build.scala
+++ b/build/build.scala
@@ -19,6 +19,17 @@ class Build(context: Context) extends BasicBuild(context){
 
   private object tutCache extends Cache[Unit]
   def tut = {
-    println(TutMain)
+
+    // tut is normally run as a commandline app so it wants strings :-\
+    // we can modify tut to add a typed runner and provide a logging hook
+    // but as a proof of concept yep it works
+    val in  = new File(projectDirectory, "doc/src/tut")
+    val out = new File(projectDirectory, "doc/target/tut")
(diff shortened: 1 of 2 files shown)


cvogt:
finally getting to this :). Merging this into my other branch for now and then reworking it as a plugin.

build/build.scala
```diff
@@ -19,6 +19,17 @@ class Build(context: Context) extends BasicBuild(context){
...
   def tut = {
-    println(TutMain)
+
+    // tut is normally run as a commandline app so it wants strings :-\
+    // we can modify tut to add a typed runner and provide a logging hook
+    // but as a proof of concept yep it works
+    val in  = new File(projectDirectory, "doc/src/tut")
+    val out = new File(projectDirectory, "doc/target/tut")
+    val re  = """.*\.(md|txt|htm|html)"""
+    val cp  = dependencyClasspath.string // we really want cbt on here too
```
cvogt: > we really want cbt on here too
why?

ShaneDelmore:
This is great. Tut was a much needed plugin for my usage. 👍

PR #118, opened 2016-05-02
Chris

cvogt:
tests failing here, because this change apparently exposes a previously existing classloading related bug in build composition.


cvogt:
merging this for my Scalapeno talk. Fixing the (previously existing) bug the new tests approach later. Meaning one test will now fail in master.

PR #126, opened 2016-05-21
Interactive

mchav:
Interactive mode with command history.

### DIFF
--- a/cbt
+++ b/cbt
@@ -225,14 +225,32 @@ stage1 () {
 }
 
 while true; do
-	stage1 $*
+	interactive=false
+
+	if [ "$1" = "interactive" ]; then
+		interactive=true
+	else
+		stage1 $*
+	fi
+
+	while [ "$interactive" = true ]; do
+		while IFS="" read -r -e -d $'\n' -p 'cbt> ' options; do
+		  history -s "$options"
+			if [ "$options" = "quit" ]; then
+				interactive=false
+				break
+			else
+				stage1 "$options"
+			fi
+		done
+	done
 	if [ ! "$1" = "loop" ]; then
 		break
 	fi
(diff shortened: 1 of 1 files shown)


cbt
```diff
@@ -225,14 +225,32 @@ stage1 () {
...
+		done
+	done
 	if [ ! "$1" = "loop" ]; then
 		break
 	fi
 	echo "======= Restarting CBT =======" 1>&2
 done
 
-if [ $compiles -ne 0 ]; then
+if [ ! "$compiles" = "0" ]; then
```
cvogt: why was this needed?

cbt
```diff
@@ -225,14 +225,32 @@ stage1 () {
...
+		while IFS="" read -r -e -d $'\n' -p 'cbt> ' options; do
+		  history -s "$options"
+			if [ "$options" = "quit" ]; then
+				interactive=false
+				break
+			else
+				stage1 "$options"
+			fi
+		done
+	done
```
cvogt: interesting :)

cvogt:
This is a quite interesting attempt. I am not convinced of this yet though. `loop` does not work and there is no history and also no access to shell commands, completion etc. I wonder if there is a way to basically transparently layer cbt commands on top of regular shell commands in the users default prompt in order not to loose the entire power of the users shell, but also not having to type `cbt` all the time. Maybe by putting CBT as a man in the middle between the user and the actual shell and running existing commands through cbt, others through the shell. Closing this for now (hope that's ok). We can always get it back: eb71e84e4361cf4abb35aa915f84889b1cc040c5

cvogt:
I stand corrected. the implementation was smarter than I thought. so it does have a history. Let's re-evaluate this-

cvogt:
closed in favor of #141

PR #127, opened 2016-05-22
Support for private git repositories.

mchav:


stage2/GitDependency.scala
```diff
@@ -20,17 +21,34 @@ case class GitDependency(
...
+          val userNameAndPassword = scala.io.Source.fromFile(context.projectDirectory + "/git.login").mkString.split("\n")
+          Some(userNameAndPassword)
+        } catch {
+          case e: FileNotFoundException => None
+        }
+      }
       val git =
         Git.cloneRepository()
           .setURI(url)
+          .setCredentialsProvider( mkCredentials( authList ) )
```
cvogt: I think nicer would be calling .setCredentialProvider only if there were credentialy, so we don't have to do the dummy stuff up there

stage2/GitDependency.scala
```diff
@@ -20,17 +21,34 @@ case class GitDependency(
...
+      case None => new UsernamePasswordCredentialsProvider("dummy", "dummy")
+    }
     val checkoutDirectory = context.cache ++ s"/git/$domain/$path/$ref"
     if(checkoutDirectory.exists){
       logger.git(s"Found existing checkout of $url#$ref in $checkoutDirectory")
     } else {
       logger.git(s"Cloning $url into $checkoutDirectory")
+      val authList = {
+        try {
+          val userNameAndPassword = scala.io.Source.fromFile(context.projectDirectory + "/git.login").mkString.split("\n")
```
cvogt: `git.login` should be on the .gitignore list. It's fine for now, but doesn't seem scalable. Different repositories may have different credentials

mchav: The gitignore has *.login so it's already there. Looking more into it.

mchav: @cvogt it seems the only way this can happen at scale is either reading from .gitconfig, or getting CLI input. A text file works for the general case though.

stage2/GitDependency.scala
```diff
@@ -17,20 +18,33 @@ case class GitDependency(
...
+        try {
+          val credentials = scala.io.Source.fromFile(context.projectDirectory + "/git.login").mkString.split("\n")
+          Some(GitCredentials(credentials(0), credentials(1)))
+        } catch {
+          case e: FileNotFoundException => None
+        }
+      }
+      val git = authenticate(credentials, Git.cloneRepository().setURI(url))
           .setDirectory(checkoutDirectory)
           .call()
```
cvogt: What happens if the repository needed credentials, but we didn't provide any?

cvogt:
superseded by #214

PR #131, opened 2016-05-27
Fixing scalatest plugin + example folder with scalatest sample build

katrinsharp:


examples/build-scalatest/src/test/scala/Test.scala
```diff
@@ -0,0 +1,24 @@
...
+    val emptyStack = new Stack[Int]
+    a [NoSuchElementException] should be thrownBy {
+      emptyStack.pop()
+    }
+  }
+
+  "square" should "return double" in {
+    Main.square(2) should be (4)
+  }
+}
```
cvogt: Not super important, but let's try to end files on new-lines to adhere to posix standard (http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206). This will come in handy should anyone ever try to use unix tools like sed on the source code. I have seen sed automatically add newlines to the ends of all files for example, probably assuming that would not change anything, but resulting in lots of files showing up as changed.

stage2/BasicBuild.scala
```diff
@@ -54,7 +54,7 @@ class BasicBuild(val context: Context) extends DependencyImplementation with Bui
     )
 
   // ========== paths ==========
-  final private val defaultSourceDirectory = projectDirectory ++ "/src"
+  protected def defaultSourceDirectory = projectDirectory ++ "/src"
```
cvogt: do we need to make this non-private? I was expecting people to override def sources instead. Any reason against that?

examples/build-scalatest/build/build/build.scala
```diff
@@ -0,0 +1,14 @@
...
+import java.net.URL
+import java.io.File
+import scala.collection.immutable.Seq
+
+class Build( context: Context ) extends BuildBuild( context ){
+
+  override def dependencies = (
+    super.dependencies ++
+      Resolver( mavenCentral ).bind(
+      ScalaDependency("org.scalatest","scalatest","2.2.4")
```
cvogt: the scalatest dependency is probably not needed here, right?

examples/build-scalatest/src/test/scala/Test.scala
katrinsharp: accepted

stage2/BasicBuild.scala
katrinsharp: The issue here is that projectDirectory = src/test/scala in case of ScalaTest. If we don't override defaultSourceDirectory in this case it becomes invalid src/test/scala/src. Other option will be to start restructure it so ScalaTest case can be accommodated.

plugins/scalatest/ScalaTest.scala
```diff
@@ -16,8 +16,10 @@ trait SbtLayout extends BasicBuild{
   override def sources = Seq( projectDirectory ++ "/src/main/scala" )
   def testSources = projectDirectory ++ "/src/test/scala"
   def testDependencies: Seq[Dependency] = Nil
-  lazy val testBuild = 
+  lazy val testBuild =
     new BasicBuild(context.copy(projectDirectory = testSources)) with ScalaTest{
+      override def defaultSourceDirectory = projectDirectory
+      override def sources = Seq(defaultSourceDirectory)
```
cvogt: what if we only `override def sources = Seq(projectDirectory)` here?

stage2/BasicBuild.scala
cvogt: commented above

katrinsharp:
On bright note, I've added testing example build to circle, so we can't easily break things.

plugins/scalatest/ScalaTest.scala
cvogt: or better: `override def sources = Seq(testSources)`
and set the project directory somewhere else, so that the target folder does not end up in `/src/test/scala`

cvogt:
@katrinsharp yeah, great stuff. Super happy to see you jump on the ScalaTest topic. I really just hacked something working together here, it totally needs more love.

plugins/scalatest/ScalaTest.scala
cvogt: actually nevermind about the target folder. we reset this below. Not sure if the projectDirectory is all that important here now.

katrinsharp: We could check if projectDirectory ends with 'src' and if it is, then not to append - I don't like that idea as much. Other option would be to leave defaultSourceDirectory to what it does today in master, but remove the assertion of assertSourceDirectories(). Not the best thing either. The third option not to add 'src' to defaultSourceDirectory unless asked by some sort of configuration/booleab, something like that.

cvogt: what if we just leave `defaultSourceDirectory` untouched but ignore it here and simply `override def sources = Seq(testSources)`. `assertSourceDirectories()` should work fine then.

cvogt: I agree that conditionally adding `src` is not ideal. would like to avoid that

katrinsharp: Ok, works, accepted.

katrinsharp:
@cvogt The CI failure happens due to the fact that my branch doesn't have #130  merged. Code related to this PR works. Should we try to merge or I should recreate PR?

stage2/BasicBuild.scala
```diff
@@ -54,7 +54,7 @@ class BasicBuild(val context: Context) extends DependencyImplementation with Bui
     )
 
   // ========== paths ==========
-  final private val defaultSourceDirectory = projectDirectory ++ "/src"
+  private final val defaultSourceDirectory = projectDirectory ++ "/src"
```
cvogt: mind undoing that?

cvogt:
@katrinsharp actually there is one failing test in master which I still need to fix. It's an abstract method error. as long as that's the only one we are good for a merge. Looks like the later commits were just refinements of the first. Mind squashing them?

stage2/BasicBuild.scala
katrinsharp: of course not. Was thinking if I'm getting it same direction ..

katrinsharp:
Goes without saying :)

katrinsharp:
@cvogt done

cvogt:
awesome. Will check test results tomorrow and merge

examples/build-scalatest/build/build.scala
```diff
@@ -0,0 +1,14 @@
...
+import java.net.URL
+import java.io.File
+import scala.collection.immutable.Seq
+
+class Build( context: Context ) extends BasicBuild( context ) with SbtLayout {
+
+  override def dependencies = (
+    super.dependencies ++
+      Resolver( mavenCentral ).bind(
+        ScalaDependency("org.scalatest","scalatest","2.2.4")
```
cvogt: does this need the scalatest dependency?

katrinsharp: Yes. BuildBuild needs it for plugin functioning, Build needs it for actual tests that use all ScalaTest functionality.

Issue #137, opened 2016-06-03
remove `Context` boiler plate from build files


cvogt:
In order to remove noise from build files we could have Context be concrete, but undefined on the BasicBuild and then inject it later

```scala
class BasicBuild{
  def context: Context = throw new Exception("Context is injected by CBT via refinement")
}
```

A concrete build.scala would then simply look like this:
```scala
class Build extends BasicBuild{ ... }
```

instead of this
```scala
class Build(context: Context) extends BasicBuild(context){ ... }
```

cbt would have another source file that it would compile together with the user provided source files for every build:

```scala
class BuildWithContext(override val context: Context) extends Build
```

cbt would then instantiate `BuildWithContext` instead of `Build`as it is now.

We could also make `Build` a trait, but this would obfuscate compile errors in case of unimplemented abstract members. They would show as missing in `BuildWithContext` instead of missing in`Build` where they make more sense for the user.

(cc @densh)

nafg:
Why not make it abstract?

cvogt:
@nafg same problem like trait. If the parent class requires implementing a field that the user forgets, the error message the users sees would be something like `class BuildWithContext in file /path/to/cbt/with_context/BuildWithContext.scala:3 does not implement required member foo`. More helpful would be `class /path/to/user/build/build.scala:3 does not implement required member foo`.

cvogt:
I am actually happy with the current state
```class Build(val context: Context) extends BaseBuild```
@nafg supports trying again to switch to traits
```trait Build extends BaseBuild```
to reduce boiler plate

Seems harder to implement at the moment. Putting this into the backlog unless somebody steps up. If nobody does, we may want to close this eventually.

cvogt:
@densh scheduled allowing `trait Build` instead of `class Build(val: context: Context)` this milestone 1.0 :) cc @nafg

densh:
@cvogt Lookimg forward to it!

PR #139, opened 2016-06-04
Reduce boiler plate from build files by aliasing File, URL, Seq in BasicBuild

cvogt:
Discussed here on twitter https://gitter.im/cvogt/cbt?at=5752127e1097267921302d84

This might actually be a bad idea. While it does reduce the boiler plate of build files, it adds magic (if you count injection via inheritance scope as magical, which is probably a good idea).

Somewhat related to #137.

### DIFF
--- a/build/build.scala
+++ b/build/build.scala
@@ -1,8 +1,4 @@
 import cbt._
-import java.net.URL
-import java.io.File
-import scala.collection.immutable.Seq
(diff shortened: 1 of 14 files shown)


cvogt:
closing this for now. general agreement on gitter seemed to be that this makes things more complicated

PR #141, opened 2016-06-07
Interactive mode implemented with readline. This allows not having to…

cvogt:
… tpe "cbt " all the time. It's not as powerful as your native shell, but at least has a command history. When you type a command into cbt's shell, it tries to run system commands first and only if none was found tries to run a cbt task for the provided name. This is an experimental design trying to save people from some constant typing pain. (implemented by @mchav, commit message by @cvogt)

### DIFF
--- a/cbt
+++ b/cbt
@@ -225,10 +225,39 @@ stage1 () {
 }
 
 while true; do
-	stage1 $*
+	interactive=false
+
+	if [ "$1" = "interactive" ]; then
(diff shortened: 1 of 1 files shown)


cvogt:
after discussing with @mchav we decided to remove this from cbt and make this a separate tool instead that could be used with any command that has subcommands like cbt, git, svn. It's purely saves typing the prefix, nothing else.

PR #144, opened 2016-06-13
scalajs cross project support as plugin

katrinsharp:
Cross project support + example build with running example showing how to use it. Please 'missing features' in README for list of what missing. Counterparts of all missing features reside in Scalajs SBT plugin. We will need to port the associated code here. I prefer to move with this pull request first since it enables working scalajs build as shown in build example.

### DIFF
--- a/examples/build-scalajs/build/build.scala
+++ b/examples/build-scalajs/build/build.scala
@@ -1,0 +1,27 @@
+import cbt._
+import java.net.URL
+import java.io.File
+import scala.collection.immutable.Seq
+
+class Build( context: Context ) extends BasicBuild( context ) with ScalaJsBuild {
+
+  override val projectName = "my-project"
+
+  override def dependencies = (
+    super.dependencies ++
+      Resolver( mavenCentral ).bind(
+        //"org.scalatest" %%% "scalatest" % "3.0.0-RC2",
+        "com.github.japgolly.scalajs-react" %%% "core" % "0.10.4", // for example
+        // for example if you want explicitely state scala version
+        "org.scala-js" % "scalajs-dom_sjs0.6_2.11" % "0.9.0"
+      )
+  )
+
+  /* ++ some JVM only dependencies */
+  override def jvmDependencies = Seq.empty
+
+  override def fastOptOutput = {
+    projectDirectory.getAbsolutePath + "/server/public/" + new File(super.fastOptOutput).getName
+  }
+}
+
--- a/plugins/scalajs/ScalaJs.scala
+++ b/plugins/scalajs/ScalaJs.scala
@@ -1,0 +1,82 @@
+import java.io.File
+
+import cbt._
+
+import scala.collection.immutable.Seq
+
+trait ScalaJsSbtDependencyDsl extends SbtDependencyDsl { self: ScalaJsBuild =>
+
+  //Note: We make same assumption about scala version.
+  //In order to be able to choose different scala version, one has to use %.
+  implicit class ScalaJsDependencyBuilder(groupId: String){
+    def %%%(artifactId: String) = new DependencyBuilder2(
+      groupId, artifactId + artifactIdSuffix, Some(scalaMajorVersion))
+  }
+}
+
+trait ScalaJsBuild extends BasicBuild with ScalaJsSbtDependencyDsl with ScalaJsInformation { outer =>
+
+  def sharedFolder = projectDirectory ++ "/shared"
+  def jvmFolder = projectDirectory ++ "/jvm"
+  def jsFolder = projectDirectory ++ "/js"
+
+  private lazy val jvmBuild = new BasicBuild(outer.context){
+    override def sources = Seq(sharedFolder ++ "/src/main/scala", jvmFolder ++ "/src/main/scala")
+    override def target = jvmFolder ++ "/target"
+    override def dependencies = outer.dependencies ++ jvmDependencies
+  }
+  private lazy val jsBuild = new BasicBuild(outer.context){
+    override def sources = Seq(sharedFolder ++ "/src/main/scala", jsFolder ++ "/src/main/scala")
+    override def target = jsFolder ++ "/target"
+    override def dependencies = outer.dependencies :+ scalaJsLibDep
+    override def scalacOptions = super.scalacOptions ++
+      Seq(s"-Xplugin:${scalaJsCompilerDep.jar.getAbsolutePath}", "-Xplugin-require:scalajs")
+  }
+
+  def jvmDependencies = Seq.empty[Dependency]
+  //TODO: implement
+  def jsDependencies = Seq.empty[Dependency]
+  def jvmCompile: Option[File] = jvmBuild.compile
+  def jsCompile: Option[File] = jsBuild.compile
+  override def compile = {
+    jvmCompile
+    jsCompile
+  }
+
+  trait JsOutputMode {
+    def option: String
+    def fileSuffix: String
+  }
+  case object FastOptJS extends JsOutputMode{
+    override val option = "-f"
+    override val fileSuffix = "fastopt"
+  }
+  case object FullOptJS extends JsOutputMode{
+    override val option = "-u"
+    override val fileSuffix = "fullopt"
+  }
+
+  private def output(mode: JsOutputMode) = s"${jsBuild.target.getAbsolutePath}/$projectName-${mode.fileSuffix}.js"
+
+  //TODO: should process all options that Scalajsld recognizes?
+  private def link(mode: JsOutputMode, outputPath: String) = {
+    lib.runMain(
+      "org.scalajs.cli.Scalajsld",
+      Seq(
+        mode.option,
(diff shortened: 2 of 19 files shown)


test/simple-fixed/build/build.scala
```diff
@@ -26,7 +26,7 @@ class Build(context: cbt.Context) extends BasicBuild(context){
       ).resolve(
         "org.cvogt" %% "play-json-extensions" % "0.8.0",
         "org.tpolecat" %% "tut-core" % "0.4.2",
-        "ai.x" %% "lens" % "1.0.0-SNAPSHOT"
+        "ai.x" %% "lens" % "1.0.0"
```
cvogt: I actually wanted to test resolving snapshots. But I have seen the problem of this not being able to be resoled, too. Does Sonatype somehow delete snapshots? When I re-deploy lens 1.0.0-SNAPSHOT it always works, but seems to stop working later.

examples/build-scalajs/build/build.scala
```diff
@@ -0,0 +1,26 @@
...
+      )
+  )
+
+  /* ++ some JVM only dependencies */
+  override def jvmDependencies = Seq.empty
+
+  override def fastOptOutput = {
+    projectDirectory.getAbsolutePath + "/server/public/" + new File(super.fastOptOutput).getName
+  }
+}
```
cvogt: \n

plugins/scalajs/ScalaJsInformation.scala
```diff
@@ -0,0 +1,24 @@
...
+  final protected val scalaJsLibDep =
+    Resolver( mavenCentral ).bindOne(
+      ScalaDependency("org.scala-js", "scalajs-library", sjsVersion)
+    )
+
+  final protected val scalaJsCliDep =
+    Resolver( mavenCentral ).bindOne(
+      ScalaDependency("org.scala-js", "scalajs-cli", sjsVersion)
+    )
+}
```
cvogt: \n

plugins/scalajs/build/build.scala
```diff
@@ -0,0 +1,8 @@
+import cbt._
+
+class Build(context: Context) extends BasicBuild(context) {
+
+  override def dependencies =
+    super.dependencies :+
+      context.cbtDependency
+}
```
cvogt: \n

cvogt:
This is really awesome! I'll take a closer look later.

katrinsharp:
FYI: @sjrd

sjrd:
How do I create a Scala.js-*only* project?

katrinsharp:
You populate js folder only. https://github.com/katrinsharp/cbt/tree/master/examples/build-scalajs has Scala.js code only and one shared model for example, that could reside in js.

sjrd:
That's weird. I can create a JVM-only project in a "normal" way, but if I want a JS-only project, I have to create a cross-project and ignore the JVM part?

katrinsharp:
Do you know how popular Scala.js-only projects are vs cross project? People usually love Scala.js for its capabilities to have shared parts between Javascript and JVM on top of not having write javascript :). I understand your concern of being weird so how about treating this PR as support cross project only and creating another issue for supporting Scala.js-only projects and let people vote? I prefer to add test capabilities to cross project as higher priority than supporting Scala.js-only.

plugins/scalajs/ScalaJs.scala
```diff
@@ -0,0 +1,82 @@
...
+        mode.option,
+        "-s",
+        "--stdlib", s"${scalaJsLibDep.jar.getAbsolutePath}",
+        "-o", outputPath,
+        jsBuild.target.getAbsolutePath) ++
+        jsBuild.dependencies.collect{case d: BoundMavenDependency => d.jar.getAbsolutePath},
+      scalaJsCliDep.classLoader(jsBuild.context.classLoaderCache))
+  }
+  def fastOptJS = link(FastOptJS, fastOptOutput)
+  def fullOptJS = link(FullOptJS, fullOptOutput)
```
cvogt: Those two should probably call `compile` first. Then we can remove it from the docs.

plugins/scalajs/ScalaJs.scala
```diff
@@ -0,0 +1,82 @@
...
+    override def dependencies = outer.dependencies ++ jvmDependencies
+  }
+  private lazy val jsBuild = new BasicBuild(outer.context){
+    override def sources = Seq(sharedFolder ++ "/src/main/scala", jsFolder ++ "/src/main/scala")
+    override def target = jsFolder ++ "/target"
+    override def dependencies = outer.dependencies :+ scalaJsLibDep
+    override def scalacOptions = super.scalacOptions ++
+      Seq(s"-Xplugin:${scalaJsCompilerDep.jar.getAbsolutePath}", "-Xplugin-require:scalajs")
+  }
+
```
cvogt: we should add `override def triggerLoopFiles = super.triggerLoopFiles ++ (jvmBuild.sources ++ jsBuild.sources).distinct` here. Then `cbt loop fastOptJs` works if `fastOpsJs` also calls `compile`.

plugins/scalajs/ScalaJs.scala
```diff
@@ -0,0 +1,82 @@
...
+  def jvmDependencies = Seq.empty[Dependency]
+  //TODO: implement
+  def jsDependencies = Seq.empty[Dependency]
+  def jvmCompile: Option[File] = jvmBuild.compile
+  def jsCompile: Option[File] = jsBuild.compile
+  override def compile = {
+    jvmCompile
+    jsCompile
+  }
```
cvogt: This sounds like this build should simply depend on jvmBuild and jsBuild instead of delegating `compile` explicitly

plugins/scalajs/ScalaJs.scala
```diff
@@ -0,0 +1,82 @@
...
+      "org.scalajs.cli.Scalajsld",
+      Seq(
+        mode.option,
+        "-s",
+        "--stdlib", s"${scalaJsLibDep.jar.getAbsolutePath}",
+        "-o", outputPath,
+        jsBuild.target.getAbsolutePath) ++
+        jsBuild.dependencies.collect{case d: BoundMavenDependency => d.jar.getAbsolutePath},
+      scalaJsCliDep.classLoader(jsBuild.context.classLoaderCache))
+  }
```
cvogt: This should probably be in a separate lib object, so that we don't tie ScalaJS to cbt's OO api. How about `object ScalaJsLib` ?

plugins/scalajs/ScalaJs.scala
```diff
@@ -0,0 +1,82 @@
...
+  trait JsOutputMode {
+    def option: String
+    def fileSuffix: String
+  }
+  case object FastOptJS extends JsOutputMode{
+    override val option = "-f"
+    override val fileSuffix = "fastopt"
+  }
+  case object FullOptJS extends JsOutputMode{
+    override val option = "-u"
```
cvogt: how about we use the full flags here, not the acronyms :)?

plugins/scalajs/ScalaJs.scala
```diff
@@ -0,0 +1,82 @@
...
+    override val fileSuffix = "fullopt"
+  }
+
+  private def output(mode: JsOutputMode) = s"${jsBuild.target.getAbsolutePath}/$projectName-${mode.fileSuffix}.js"
+
+  //TODO: should process all options that Scalajsld recognizes?
+  private def link(mode: JsOutputMode, outputPath: String) = {
```
cvogt: we could allow all options, by just taking a `scalaJsOptions: Seq[String]` as a argument and concatting it into the args

stage2/BasicBuild.scala
```diff
@@ -79,17 +80,14 @@ class BasicBuild(val context: Context) extends DependencyImplementation with Bui
...
+  protected def logEmptySourceDirectories(): Unit = {
     val nonExisting =
       sources
         .filterNot( _.exists )
         .diff( Seq(defaultSourceDirectory) )
-    assert(
-      nonExisting.isEmpty,
-      "Some sources do not exist: \n"++nonExisting.mkString("\n")
-    )
+    if(!nonExisting.isEmpty) logger.stage2("Some sources do not exist: \n"++nonExisting.mkString("\n"))
```
cvogt: good idea!

cvogt:
Merging this as is and addressing the mentioned things in a separate PR myself. I think I can also take care of the issue @sjrd mentioned.

plugins/scalajs/ScalaJs.scala
katrinsharp: I just tried to follow the same user experience of sbt Scalajs, but it sounds good to have such a dependency.

plugins/scalajs/ScalaJs.scala
katrinsharp: Does it mean that use will have to be aware/maintain jvm and js builds in their projects? Was on the fence for this one but once again meanwhile followed current sbt based user experience

PR #151, opened 2016-06-19
Scalariform support

### LINKED ISSUE #40: scalariform
write a plugin for it

rockjam:
fixes #40

### DIFF
--- a/plugins/scalariform/Scalariform.scala
+++ b/plugins/scalariform/Scalariform.scala
@@ -1,0 +1,62 @@
+package cbt
+
+import java.io.File
+import java.nio.file.FileSystems
+import java.nio.file.Files._
+
+import scala.util.Try
+import scalariform.formatter.ScalaFormatter
+import scalariform.formatter.preferences.FormattingPreferences
+import scalariform.parser.ScalaParserException
+
+trait Scalariform extends BaseBuild {
+  def scalariformFormat: ExitCode = {
+    Scalariform.format(sourceFiles, scalariformPreferences, scalaVersion)
+    ExitCode.Success
+  }
+
+  def scalariformPreferences: FormattingPreferences = Scalariform.defaultPreferences
+}
+
+object Scalariform {
+
+  val defaultPreferences: FormattingPreferences = {
+    import scalariform.formatter.preferences._
+    FormattingPreferences()
+      .setPreference(AlignParameters, true)
+      .setPreference(AlignArguments, true)
+      .setPreference(AlignSingleLineCaseStatements, true)
+      .setPreference(MultilineScaladocCommentsStartOnFirstLine, true)
+      .setPreference(SpaceInsideParentheses, true)
+      .setPreference(SpacesWithinPatternBinders, true)
+      .setPreference(SpacesAroundMultiImports, true)
+      .setPreference(DoubleIndentClassDeclaration, false)
+  }
+
+  private val scalaFileMatcher = FileSystems.getDefault.getPathMatcher("glob:**.scala")
+
+  def format(files: Seq[File], preferences: FormattingPreferences, scalaVersion: String) = {
+    var reformattedCount: Int = 0
+    for (file <- files if file.exists) {
+      val path = file.toPath
+      if(scalaFileMatcher.matches(path)) {
+        try {
+          val sourceCode = new String(readAllBytes(path))
+          val formatted = ScalaFormatter.format(
+            sourceCode,
+            preferences,
+            Some(scalaVersion)
+          )
+          if (sourceCode != formatted) {
+            write(path, formatted.getBytes)
+            reformattedCount += 1
+          }
+        } catch {
+          case e: ScalaParserException => System.err.println(s"Scalariform parser error: ${e.getMessage} when formatting source: $file")
+        }
+      }
+    }
+    System.err.println(s"Reformatted $reformattedCount Scala sources")
+  }
+
+}
--- a/examples/scalariform-example/README.md
+++ b/examples/scalariform-example/README.md
@@ -1,0 +1,5 @@
+This example shows integration with scalariform plugin.
+Reformat executed on every `cbt compile` call, and affects only *.scala source files.
+You can provide your custom scalariform preferences in build via `scalariformPreferences`.
+To test formatting in action you can execute: `cbt breakFormatting` to break formatting
+and `cbt scalariformReformat` to get formatting back.
--- a/examples/scalariform-example/build/build.scala
+++ b/examples/scalariform-example/build/build.scala
@@ -1,0 +1,31 @@
+import cbt._
+import scalariform.formatter.preferences._
+
+class Build(val context: Context) extends BuildBuild with Scalariform {
+  override def compile = {
+    scalariformFormat
+    super.compile
+  }
+
+  override def scalariformPreferences =
+    FormattingPreferences()
+      .setPreference(SpacesAroundMultiImports, true)
+      .setPreference(DoubleIndentClassDeclaration, true)
+      .setPreference(RewriteArrowSymbols, true)
+
+  def breakFormatting = {
+    import java.nio.file._
+    import scala.collection.JavaConverters._
+    sourceFiles foreach { file =>
+      try {
+        val path = file.toPath
+        val fileLines = Files.readAllLines(path).asScala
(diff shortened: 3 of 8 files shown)


rockjam:
This pull request contains basic scalariform plugin implementation without any caching.
ffbeb84 is purely related to reformatting source code of cbt with preferences from [here](https://github.com/cvogt/cbt/issues/40#issuecomment-193024198)

rockjam:
CI fails to execute `./cbt compile` with these changes, whereas it's fine on my local machine. Could this possibly happen cause of build file name? It is Build.scala file with `Build` class in it. https://github.com/rockjam/cbt/blob/cac21221a77610842b59de8462e547960e409aeb/build/build/Build.scala

cvogt:
wow, this is awesome! I'll checkout why it fails.

build/build/Build.scala
```diff
@@ -0,0 +1,5 @@
+import cbt._
+
+class Build(val context: Context) extends BuildBuild {
+  override def dependencies = super.dependencies ++ Seq(plugins.scalariform)
+}
```
cvogt: Not super important, but let's try to end files on new-lines to adhere to posix standard (http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206). This will come in handy should anyone ever try to use unix tools like sed on the source code. I have seen sed automatically add newlines to the ends of all files for example, probably assuming that would not change anything, but resulting in lots of files showing up as changed.

plugins/scalariform/Scalariform.scala
```diff
@@ -0,0 +1,63 @@
+import cbt._
```
cvogt: let's put this into package cbt instead, so it's not top-level to avoid potential conflicts with user written classes/objects.

plugins/scalariform/Scalariform.scala
```diff
@@ -0,0 +1,63 @@
...
+import java.nio.file.FileSystems
+import java.nio.file.Files._
+
+import scala.util.Try
+import scalariform.formatter.ScalaFormatter
+import scalariform.formatter.preferences.FormattingPreferences
+import scalariform.parser.ScalaParserException
+
+trait Scalariform extends BaseBuild {
+  def scalariformReformat: ExitCode = {
```
cvogt: I think the sbt-scalariform task is called `scalariformFormat`. Let's use that.

plugins/scalariform/Scalariform.scala
```diff
@@ -0,0 +1,63 @@
...
+    Scalariform(sourceFiles, scalariformPreferences, scalaVersion)
+    ExitCode.Success
+  }
+
+  def scalariformPreferences: FormattingPreferences = Scalariform.defaultPreferences
+
+  override def dependencies = super.dependencies ++ Resolver(mavenCentral).bind(ScalaDependency("org.scalariform", "scalariform", "0.1.8"))
```
cvogt: This adds scalariform as a dependency to the project that's formatted. Doesn't seem right. We can probably just delete this line.

build/build.scala
```diff
@@ -1,6 +1,6 @@
 import cbt._
 
-class Build(val context: Context) extends BaseBuild{
+class Build(val context: Context) extends BaseBuild with Scalariform {
```
cvogt: Let's not add this to CBT's build itself quite yet. Happy to do it at some point, but would like to merge some un-merged stuff first without conflict. Btw, this build file isn't really used right now. CBT builds itself with an internal bootstrapping mechanism, not with CBT. I only added this build file as a starting point to eventually deploy to maven and generate `tut` documentation. Scalariform would fit right in here as well, but at the moment this build file is not called during the normal workflow.

plugins/scalariform/Scalariform.scala
```diff
@@ -0,0 +1,63 @@
...
+      .setPreference(MultilineScaladocCommentsStartOnFirstLine, true)
+      .setPreference(SpaceInsideParentheses, true)
+      .setPreference(SpacesWithinPatternBinders, true)
+      .setPreference(SpacesAroundMultiImports, true)
+      .setPreference(DoubleIndentClassDeclaration, false)
+  }
+
+  private val scalaFileMatcher = FileSystems.getDefault.getPathMatcher("glob:**.scala")
+
+  def apply(files: Seq[File], preferences: FormattingPreferences, scalaVersion: String) = {
```
cvogt: should we call this `def format`? Otherwise it looks like a constructor.

plugins/scalariform/Scalariform.scala
```diff
@@ -0,0 +1,63 @@
...
+          if (sourceCode != formatted) {
+            write(path, formatted.getBytes)
+            reformattedCount += 1
+          }
+        } catch {
+          case e: ScalaParserException => System.err.println(s"Scalariform parser error: ${e.getMessage} when formatting source: $file")
+        }
+      }
+    }
+    println(s"Reformatted $reformattedCount Scala sources")
```
cvogt: Let's print things like this using System.err.println. So far I've been using System.out for machine-readable info and System.err for debug output that should be shown to the user by default.

cvogt: Otherwise ```scala -cp `cbt classpath` ``` would break for example if the compile task is set to depend on the format task, because it would not return not only the classpath, but also this message

examples/scalariform-example/src/main/scala/Main.scala
```diff
@@ -0,0 +1,10 @@
+//This source file will be reformated after `cbt compile`/`cbt run`
+import scala.concurrent.{Future,Await}
+
+object Main extends App{
+println("fooo")
+List(1, 2, 4, 5, 6) match {
+case h :: _ => println("not empty list")
+case Nil => println("empty list")
+}
+}
```
cvogt: Ideally we should be able to run scalariform during testing. This would always leave this file changed. Should be maybe check it into git already formatted, but have the tests remove spaces in front of the newlines, then run and make sure the files is back to normal?

cvogt:
I just wrote out my thoughts in comments. Let me know what you think. Happy to help with any changes.

examples/scalariform-example/src/main/scala/Main.scala
cvogt: oh also, with cbt, there is no need for `src/main/scala/` You can put them into the top-level directory or anywhere under `src/` works. I know people are used to it, but I think it's time to start cutting ties to old baggage and steer towards a new, cleaner future :).

examples/scalariform-example/src/main/resources/reference.conf
```diff
@@ -0,0 +1,8 @@
+// should not reformat this, cause it is not in source files
+some {
+  inside {
+    foo: 22
+    bar: false
+    baz: "hello"
+  }
+}
```
cvogt: Not important here, but CBT is currently unaware of `src/main/resources/`. We should probably tell it to add them to the classpath eventually. And probably a good place to put them is just `resources/` in the top-level directory

examples/scalariform-example/build/build/Build.scala
```diff
@@ -0,0 +1,5 @@
+import cbt._
+
+class Build(val context: Context) extends BuildBuild {
+  override def dependencies = super.dependencies ++ Seq(plugins.scalariform )
```
cvogt: doesn't matter, but easier than `++ Seq(...)` is `:+`

cvogt: same in examples/scalariform-example/build/Build.scala

cvogt:
tried it locally. works like a charm. Sweet :). Still looking into the build failure. It's weird.

cvogt:
@rockjam yes, so you are right. CBT currently expects the file to be `build.scala` not `Build.scala`. Under windows and OSX this should not be a problem right now, but circleci runs on linux with a case sensitive file system. I opened a ticket to fix the issue (https://circleci.com/gh/cvogt/cbt/312), but for now can you lower case your build files as I did in https://github.com/cvogt/cbt/commit/efc24c07d323412a928854d36d99122b504a9d12 ? You can just cherry-pick that one if you want.

rockjam:
@cvogt thanks for review, agree on your comments, will fix it soon

rockjam:
@cvogt updated with fixes

cvogt:
awesome stuff

PR #159, opened 2016-06-23
Uber-jar plugin initial implementation

### LINKED ISSUE #124: fat jar
we should allow generating fat jars. Probably as a plugin. It's been requested by multiple people

rockjam:
fixes #124 
Still not finished, but works already. Need to be polished and tested


rockjam:
@cvogt @mchav Here is my version of uber jar(fat jar) plugin implementation. Will add examples a bit later

plugins/uber-jar/build/build.scala
```diff
@@ -0,0 +1,3 @@
+import cbt._
+
+class Build(val context: Context) extends Plugin
```
rockjam: Not sure if we even need this. We don't right?

cvogt: We do. without it the project's code has no access to CBT's classes.

cvogt:
do people generally call this thing fatjar? Maybe we should call the plugin FatJar then

rockjam:
Some people call them "fat jar", some call them "uber jar", "uber jar" is actually common name for this

cvogt:
ok, either way works for me

cvogt:
I'll start writing documentation soonish will have a section about "built-in" plugins

rockjam:
That's great. I'll try to add examples soon, and also polish some places

rockjam:
@cvogt @mchav Updated PR, added example with README. It would be perfect to add multi-module example. I'm pretty sure it will work fine, but need to check. I think it worth to test this plugin on more examples. Would you help me on that?

rockjam:
It's actually kinda tricky. I guess plugin won't work with java sources yet, and I'm not sure that resources will be handled correctly. As far as I know, cbt itself doesn't handle resources now.

plugins/uber-jar/src/UberJar.scala
```diff
@@ -0,0 +1,128 @@
...
+      Files.deleteIfExists(path)
+      Files.createFile(path)
+      log("Creating jar file - DONE")
+      path
+    }
+    withCloseable(new JarOutputStream(new FileOutputStream(jarPath.toFile), createManifest(mainClass))) { out =>
+      writeTarget(compilerTarget, out)(log)
+
+      // it will contain all jar files, including jars, that cbt depend on! Not good!
+      val jars = classpath.files filter (f => jarFileMatcher.matches(f.toPath))
```
rockjam: Not happy about this. It will contain cbt's jar dependencies, they aren't relevant to project. What can we do with it?

plugins/uber-jar/src/UberJar.scala
```diff
@@ -0,0 +1,136 @@
...
+
+import scala.util.{Failure, Success, Try}
+
+trait UberJar extends BaseBuild {
+
+  private val log: String => Unit = logger.log("uber-jar", _)
+
+  final def uberJar: ExitCode = {
+    // we need compile first to produce target directory
+    compile
```
cvogt: no need to. `classpath` which you use below triggers that already

plugins/uber-jar/src/uberjar/JarUtils.scala
```diff
@@ -0,0 +1,105 @@
...
+      path
+    }
+    jars foreach { jar => extractJar(jar, destDir)(log) }
+    destDir
+  }
+
+  /**
+    * Extracts content of single jar file to destination directory.
+    * When extracting jar, if same file already exists, we skip(don't write) this file.
+    * TODO: maybe skipping duplicates is not best strategy. Figure out duplicate strategy.
```
cvogt: why don't we handle conflicts like this: as part of the uberJar process we let the user provide function from a list of pairs (jar file, contained file) and they have to give us back a list of pairs should be included. We throw an exception if it contains duplicates. Default implementation either `identity` (meaning we fail) or something like `_.filterNot(_._2 == "Manifest....")`. Better to fail and require solution than silently discard things during conflicts.

cvogt:
@rockjam I think we should simplify this. Here are a few ideas:

1. How about we do NOT catch the exceptions? In a way they are relatively good error messages for the user. The `uberJar` task name appears in the stack trace. They point exactly to what went wrong. Since CBT plugins are consumed as source code, they have the source code as well to look further.
2. Please check `Lib.jarFile` in `stage2/Lib.scala`. Can you simply use that or does that not work? (we should probably add a `try-finally` there for the `.close`, right?
3. regarding `withClosable` - I like the intention, but if we can reduce the places where it would be used to very few, I'd prefer `.close` for the moment I think, in order to avoid complexity by abstract. WDYT?

plugins/uber-jar/src/UberJar.scala
cvogt: sounds like a bug

examples/uber-jar-example/build/build.scala
```diff
@@ -0,0 +1,16 @@
+import cbt._
+
+class Build(val context: Context) extends BuildBuild with UberJar {
```
cvogt: Here is the problem why it includes cbt's dependencies in the fatjar. Your build is a `BuildBuild`. Should be a `BaseBuild`. Creating a ticket to catch such mistakes.

cvogt: done https://github.com/cvogt/cbt/issues/174

plugins/uber-jar/src/UberJar.scala
cvogt: see my comment way above

rockjam: Tried `cbt classpath` again on uber-jar-example and here what I got:
```
/Users/rockjam/projects/cbt/cache/maven/com/chuusai/shapeless_2.11/2.3.1/shapeless_2.11-2.3.1.jar
/Users/rockjam/projects/cbt/cache/maven/com/github/mpilquist/simulacrum_2.11/0.7.0/simulacrum_2.11-0.7.0.jar
/Users/rockjam/projects/cbt/cache/maven/com/googlecode/javaewah/JavaEWAH/0.7.9/JavaEWAH-0.7.9.jar
/Users/rockjam/projects/cbt/cache/maven/com/jcraft/jsch/0.1.53/jsch-0.1.53.jar
/Users/rockjam/projects/cbt/cache/maven/com/lihaoyi/fansi_2.11/0.1.3/fansi_2.11-0.1.3.jar
/Users/rockjam/projects/cbt/cache/maven/com/lihaoyi/sourcecode_2.11/0.1.1/sourcecode_2.11-0.1.1.jar
/Users/rockjam/projects/cbt/cache/maven/commons-codec/commons-codec/1.6/commons-codec-1.6.jar
/Users/rockjam/projects/cbt/cache/maven/commons-logging/commons-logging/1.1.3/commons-logging-1.1.3.jar
/Users/rockjam/projects/cbt/cache/maven/net/incongru/watchservice/barbary-watchservice/1.0/barbary-watchservice-1.0.jar
/Users/rockjam/projects/cbt/cache/maven/net/java/dev/jna/jna/3.2.2/jna-3.2.2.jar
/Users/rockjam/projects/cbt/cache/maven/org/apache/httpcomponents/httpclient/4.3.6/httpclient-4.3.6.jar
/Users/rockjam/projects/cbt/cache/maven/org/apache/httpcomponents/httpcore/4.3.3/httpcore-4.3.3.jar
/Users/rockjam/projects/cbt/cache/maven/org/eclipse/jgit/org.eclipse.jgit/4.2.0.201601211800-r/org.eclipse.jgit-4.2.0.201601211800-r.jar
/Users/rockjam/projects/cbt/cache/maven/org/scala-lang/modules/scala-xml_2.11/1.0.5/scala-xml_2.11-1.0.5.jar
/Users/rockjam/projects/cbt/cache/maven/org/scala-lang/scala-library/2.11.8/scala-library-2.11.8.jar
/Users/rockjam/projects/cbt/cache/maven/org/scala-lang/scala-reflect/2.11.7/scala-reflect-2.11.7.jar
/Users/rockjam/projects/cbt/cache/maven/org/scala-sbt/test-interface/1.0/test-interface-1.0.jar
/Users/rockjam/projects/cbt/cache/maven/org/scalacheck/scalacheck_2.11/1.12.5/scalacheck_2.11-1.12.5.jar
/Users/rockjam/projects/cbt/cache/maven/org/slf4j/slf4j-api/1.7.2/slf4j-api-1.7.2.jar
/Users/rockjam/projects/cbt/cache/maven/org/typelevel/catalysts-macros_2.11/0.0.2/catalysts-macros_2.11-0.0.2.jar
/Users/rockjam/projects/cbt/cache/maven/org/typelevel/catalysts-platform_2.11/0.0.2/catalysts-platform_2.11-0.0.2.jar
/Users/rockjam/projects/cbt/cache/maven/org/typelevel/cats-core_2.11/0.6.0/cats-core_2.11-0.6.0.jar
/Users/rockjam/projects/cbt/cache/maven/org/typelevel/cats-free_2.11/0.6.0/cats-free_2.11-0.6.0.jar
/Users/rockjam/projects/cbt/cache/maven/org/typelevel/cats-kernel-laws_2.11/0.6.0/cats-kernel-laws_2.11-0.6.0.jar
/Users/rockjam/projects/cbt/cache/maven/org/typelevel/cats-kernel_2.11/0.6.0/cats-kernel_2.11-0.6.0.jar
/Users/rockjam/projects/cbt/cache/maven/org/typelevel/cats-laws_2.11/0.6.0/cats-laws_2.11-0.6.0.jar
/Users/rockjam/projects/cbt/cache/maven/org/typelevel/cats-macros_2.11/0.6.0/cats-macros_2.11-0.6.0.jar
/Users/rockjam/projects/cbt/cache/maven/org/typelevel/cats_2.11/0.6.0/cats_2.11-0.6.0.jar
/Users/rockjam/projects/cbt/cache/maven/org/typelevel/discipline_2.11/0.4/discipline_2.11-0.4.jar
/Users/rockjam/projects/cbt/cache/maven/org/typelevel/machinist_2.11/0.4.1/machinist_2.11-0.4.1.jar
/Users/rockjam/projects/cbt/cache/maven/org/typelevel/macro-compat_2.11/1.1.1/macro-compat_2.11-1.1.1.jar
/Users/rockjam/projects/cbt/compatibility/target/scala-2.11/classes/
/Users/rockjam/projects/cbt/examples/uber-jar-example/target/scala-2.11/classes/
/Users/rockjam/projects/cbt/nailgun_launcher/target/scala-2.11/classes/
/Users/rockjam/projects/cbt/stage1/target/scala-2.11/classes/
/Users/rockjam/projects/cbt/stage2/target/scala-2.11/classes/
```

There are some cbt entries, as you can see

plugins/uber-jar/src/UberJar.scala
```diff
@@ -0,0 +1,136 @@
...
+    System.err.println("Creating uber jar...")
+    UberJar.createUberJar(
+      targetDir = target,
+      compilerTarget = compileTarget,
```
cvogt: typo on the left

plugins/uber-jar/src/UberJar.scala
```diff
@@ -0,0 +1,136 @@
...
+        ExitCode.Success
+      case Failure(e) =>
+        System.err.println(s"Failed to create uber jar, cause: $e")
+        ExitCode.Failure
+    }
+  }
+
+  def uberJarMainClass: Option[String] = Some(runClass)
+
+  def uberJarName: String = projectName
```
cvogt: how about we require a `File` here instead of a string. then we don't need to pass `compileTarget` into `createUberJar`

plugins/uber-jar/src/UberJar.scala
```diff
@@ -0,0 +1,136 @@
...
+
+trait UberJar extends BaseBuild {
+
+  private val log: String => Unit = logger.log("uber-jar", _)
+
+  final def uberJar: ExitCode = {
+    // we need compile first to produce target directory
+    compile
+    System.err.println("Creating uber jar...")
+    UberJar.createUberJar(
```
cvogt: how about simply calling this `.create` to avoid the duplicate `UberJar` ?

plugins/uber-jar/src/uberjar/JarUtils.scala
```diff
@@ -0,0 +1,105 @@
...
+  protected def extractJars(jars: Seq[File])(log: String => Unit): Path = {
+    val destDir = {
+      val path = Files.createTempDirectory("unjars")
+      path.toFile.deleteOnExit()
+      log(s"Unjars directory: $path")
+      path
+    }
+    jars foreach { jar => extractJar(jar, destDir)(log) }
+    destDir
+  }
```
cvogt: so @mchav may have an opinion here. I think we actually extracts the jar into an in-memory file-system instead of disk. That may have performance benefits. Potentially it is even worth it loading the jar files into memory first, too.

plugins/uber-jar/src/uberjar/JarUtils.scala
```diff
@@ -0,0 +1,105 @@
...
+  /**
+    * If `root` is directory: writes content of directory to jar with original mapping
+    * If `root` is file: writes this file to jar
+    * @param root parent directory, with content should go to jar, or file to be written
+    * @param out jar output stream
+    * @param log logger
+    * @return returns `root`
+    */
+  protected def writeFilesToJar(root: Path, out: JarOutputStream)(log: String => Unit): Path = {
+    Files.walkFileTree(root, new SimpleFileVisitor[Path]() {
```
cvogt: see cbt's `Lib.listFilesRecursive`. it works very nicely with cbt's `Lib.jarFile` and avoids having to use a walkFileTree (which seems like a mostly archaic Java concept by now :).)

cvogt:
I suppose cbt's `Lib.listFilesRecursive` and cbt's `Lib.jarFile` could make the implementation vastly simpler. Let me know what you think.

plugins/uber-jar/src/uberjar/JarUtils.scala
rockjam: In practice, there is a lot of similar files among a lot of libraries. 
Some of them have META-INF/NOTICE.txt in , some have META-INF/LICENSE.txt, META-INF/DEPENDENCIES.
In the rest of cases `log(s"File $entryPath already exists, skipping.")` - logs directories. In case of directories we really should skip them, cause they already created.

rockjam: I think I shouldn't log when directories are duplicated. 
And I also think that we should exclude all those META-INF/<thing that common for a lot of libraries>

examples/uber-jar-example/build/build.scala
rockjam: Yep, now it's perfectly fine)

plugins/uber-jar/src/UberJar.scala
rockjam: Seems that we have to keep both. These aren't final, so it's better to keep both. What if somebody override compileTarget for some reason, and these things will be relatively different. And one more thing - I don't want people to specify full path to jar, just a name, if they want to customize it. These are my thoughts

plugins/uber-jar/src/uberjar/JarUtils.scala
cvogt: right how about we remove directories from the list already from the start, and always automatically exclude the META-INF directory?

rockjam: I agree

plugins/uber-jar/src/UberJar.scala
```diff
@@ -0,0 +1,136 @@
...
+    *  Manifest-Version: 1.0
+    *  Created-By: java.runtime.version
+    *
+    * @param mainClass optional main class
+    * @return mainifest for jar
+    */
+  private def createManifest(mainClass: Option[String]): jar.Manifest = {
+    val m = new jar.Manifest()
+    m.getMainAttributes.putValue("Manifest-Version", "1.0")
+    val createdBy = Option(System.getProperty("java.runtime.version")) getOrElse "1.7.0_06 (Oracle Corporation)"
```
cvogt: should we fail rather than using a default value here? you could simply use `sys.props("java.runtime.version")`

rockjam:
@cvogt updated PR. It actually became much simpler

plugins/uber-jar/src/UberJar.scala
```diff
@@ -0,0 +1,65 @@
...
+    val extractedJarsRoot = extractJars(jars.distinct)(log).toFile
+    log("Extracting jars - DONE")
+
+    log("Writing jar file...")
+    val optOutput = lib.jarFile(file, Seq(compileTarget, extractedJarsRoot), mainClass)
+    log("Writing jar file - DONE")
+
+    optOutput foreach { uberJar =>
+      System.err.println(s"Uber jar created. You can grab it at $uberJar")
+    }
```
cvogt: do you expect optOutput could be empty?

stage2/Lib.scala
```diff
@@ -235,7 +239,7 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
             entry.setTime(file.lastModified)
             jar.putNextEntry(entry)
             jar.write( readAllBytes( file.toPath ) )
-            jar.closeEntry
+            jar.closeEntry()
```
cvogt: Is this needed?

stage2/Lib.scala
```diff
@@ -214,15 +214,19 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
...
     } else {
       jarFile.getParentFile.mkdirs
       logger.lib("Start packaging "++jarFile.string)
-      val manifest = new Manifest
-      manifest.getMainAttributes().put(Attributes.Name.MANIFEST_VERSION, "1.0")
-      val jar = new JarOutputStream(new FileOutputStream(jarFile.toString), manifest)
+      val manifest = new Manifest()
+      manifest.getMainAttributes.put(Attributes.Name.MANIFEST_VERSION, "1.0")
+      manifest.getMainAttributes.putValue("Created-By",
+        Option(System.getProperty("java.runtime.version")) getOrElse "1.7.0_06 (Oracle Corporation)")
```
cvogt: why the getOrElse? shouldn't we rather fail if there is no java.runtime.version available?

stage2/Lib.scala
```diff
@@ -214,15 +214,19 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
...
+      mainClass foreach { className => manifest.getMainAttributes.put(Attributes.Name.MAIN_CLASS, className)
+      }
```
cvogt: can we move that the the previous line?

plugins/uber-jar/src/uberjar/JarUtils.scala
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  protected def createJarFile(parent: Path, name: String): File = {
+    val path = parent.resolve(validJarName(name))
+    Files.deleteIfExists(path)
+    Files.createFile(path)
+    path.toFile
+  }
+
+  private def validJarName(name: String) = if (name.endsWith(".jar")) name else name + ".jar"
```
cvogt: I'd prefer not to do something like this but be clear to the user if .jar should be included in the name or not. If somebody wants to call his jar   .jar.jar for some reason this would break it surprisingly

plugins/uber-jar/src/UberJar.scala
rockjam: No, I don't think it would happen

plugins/uber-jar/src/uberjar/JarUtils.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  protected val (jarFileMatcher, excludeFileMatcher) = {
+    val fs = FileSystems.getDefault
+    (fs.getPathMatcher("glob:**.jar"), fs.getPathMatcher("glob:**{.RSA,.DSA,.SF,.MF,META-INF}"))
+  }
+
+  protected def createJarFile(parent: Path, name: String): File = {
+    val path = parent.resolve(validJarName(name))
+    Files.deleteIfExists(path)
+    Files.createFile(path)
```
cvogt: let's delete the old file inside of `Lib.jarFile`. I don't think we need `createFile`. `Lib.jarFile` already creates it.

stage2/Lib.scala
rockjam: it's just a code convention for scala to put () on side-effecting function, otherwise it looks like author is trying to get a value

plugins/uber-jar/src/UberJar.scala
cvogt: ok, agreed :)

plugins/uber-jar/src/UberJar.scala
cvogt: ok it if shouldn't happen we should fail if it does, not ignore the case I think

stage2/Lib.scala
rockjam: I think we can fall in some other place. We can still create manifest, cause at the end - is't not such a big failure, from manifest's point of view.

plugins/uber-jar/src/UberJar.scala
cvogt: maybe we should guard the entire method here in `if( files.isEmpty ){`  just like jarFile. And then optOutput.getOrElse (throw new Exception("jarFile unexpectedly returned no file")) (better name for optOutput may be uberJar)

stage2/Lib.scala
rockjam: what exactly do you mean?

rockjam: like this?
```
mainClass foreach { className => 
   manifest.getMainAttributes.put(Attributes.Name.MAIN_CLASS, className)
}
```

cvogt: yeah, that's great

stage2/Lib.scala
cvogt: I generally prefer failing if something weird like that happens. People can easily path CBT and submit a PR if that's unexpected. Better than writing a wrong java runtime version. OR we write something like "Unknown". Not sure if that's valid. Or can we simply omit "Created-By" if we don't know the java runtime version?

rockjam:
@cvogt updated PR

rockjam:
@cvogt updated. now works nicely with mutli-project builds

PR #194, opened 2016-06-30
Idea plugin

prassee:
This PR implements IDEA project generator and would work for a single top level project. The implementation has some assumptions right now (will be eventually addressed):
- *.iml file will be generated by executing cbt generateIdeaProject.
- the path of cbt's comaptibility,stage1 & stage2 are injected / hardcoded in the resulting iml file.


.gitignore
```diff
@@ -12,3 +12,7 @@ realpath/realpath
 node_modules
 *fastopt*
 *fullopt*
+
+# idea specific files
+.idea/
+*.iml
```
cvogt: posix standard requires new lines at ends of files. not doing this may lead to weird behavior of tools like sed. We should try to have them.

examples/idea-plugin-example/src/Main.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import scala.concurrent.{ Await, Future }
+import scala.concurrent.duration._
+
+object Main extends App {
+  println("fooo")
+  val futureRes = Await.result(Future.successful(1), 5.seconds)
+  List(1, 2, 4, 5, 6) match {
+    case h :: _ ⇒ println("not empty list")
+    case Nil ⇒ println("empty list")
```
cvogt: I'd prefer not to have unicode arrows, but just use the ascii symbols

plugins/idea_plugin/IdeaPlugin.scala
```diff
@@ -0,0 +1,83 @@
+package idea_plugin
+
+import java.io.{File, FileWriter}
+
+import cbt.{BaseBuild, ExitCode}
+
+/**
+  * IDEA plugin
+  */
```
cvogt: Let's add comments when they add something, not otherwise :).

plugins/idea_plugin/IdeaPlugin.scala
```diff
@@ -0,0 +1,83 @@
...
+      depIdeaOrderEntry =
+      s"""<orderEntry type="module-library">
+          |      <library>
+          |        <CLASSES>
+          |            <root url="jar://$moduleDir/../cbt/${depJarFile.getPath.split("/cbt")(1)}!/" />
+          |          </CLASSES>
+          |          <JAVADOC />
+          |          <SOURCES />
+          |       </library>
+          |</orderEntry>""".stripMargin
```
cvogt: any reason not to use scala's xml syntax?

plugins/idea_plugin/IdeaPlugin.scala
```diff
@@ -0,0 +1,83 @@
...
+/**
+  * IDEA plugin
+  */
+trait IdeaPlugin extends BaseBuild {
+
+  import IdeaPlugin._
+
+  // @TODO add only the top level dependencies
+  // @TODO return exit code dynamically
+  def generateIdeaProject: ExitCode = {
```
cvogt: Let's move this into the library (and leave it publicly accessible) and only have a simple tiny method here that passes all values that are needed into the library method. That way in theory somebody could put a different frontend on top of cbt that doesn't use inheritance.

plugins/idea_plugin/build/build.scala
```diff
@@ -0,0 +1,2 @@
+import cbt._
+class Build(val context: Context) extends Plugin
```
cvogt: another missing newline

cvogt:
good stuff :)

examples/idea-plugin-example/src/Main.scala
prassee: I really don't know how it got inserted in my IDE I just see "=>"

plugins/idea_plugin/IdeaPlugin.scala
prassee: Pls explain on this further on what is that method and what parameters it might take. My guess is anyone using this trait in their build can just call cbt generateIdeaProject to create the .iml file.

cvogt: right, so let's keep generateIdeaProject, but implement it as
```
def generateIdeaProject = IdeaPlugin.generateIdeaProject( /*... whatever needed from the Build class ... */ )
```

Then someone implementing a hypothetical alternative frontend could call IdeaPlugin.generateIdeaProject(...) passing in whatever needed taken from other places.

plugins/idea_plugin/IdeaPlugin.scala
```diff
@@ -0,0 +1,85 @@
...
+
+    val imlFile = new File(projectDirectory.getPath + s"/$projectName.iml")
+    if (!imlFile.exists()) {
+      imlFile.createNewFile()
+    }
+    val fw = new FileWriter(imlFile.getPath, false)
+    fw.write(templateWithCBTSources(projectDependencies.mkString("\n"), moduleDir))
+    fw.close()
+    ExitCode.Success
+  }
```
cvogt: Mind moving most of the code in this method into a def on `IdeaPlugin` that only takes what's necessary as arguments? The reason is that I want the logic as little coupled as possible to CBT's OO layer, so that it is theoretically possible to put another layer on top that if anyone in the Scala FP community feels the necessity.

cvogt:
closing this in favor of version rebased on master and squashed: https://github.com/cvogt/cbt/pull/416

PR #196, opened 2016-07-05
Allow empty credentials for publishing.

benjaminfrank:
This enables to publish into a repository where no credentials are
required but and empty/wrong HTTP Auth header throws an HTTP 401
(not authorized) error


cvogt:
LGTM

PR #200, opened 2016-07-05
Support proxy settings in CBT.

benjaminfrank:
Proxy settings can be supplied either via Java systemproperties or via
environment variables (http_proxy / https_proxy). Java systemproperties
take precedence over env vars.
Evaluation of proxy settings happens only during nailgun startup for now.

### DIFF
--- a/nailgun_launcher/Stage0Lib.java
+++ b/nailgun_launcher/Stage0Lib.java
@@ -67,7 +67,7 @@ public class Stage0Lib{
     if(changed){
       List<String> zincArgs = new ArrayList<String>(
         Arrays.asList(
-          new String[]{ 
+          new String[]{
             "-scala-compiler", earlyDeps.scalaCompiler_2_11_8_File,
             "-scala-library", earlyDeps.scalaLibrary_2_11_8_File,
             "-scala-extra", earlyDeps.scalaReflect_2_11_8_File,
@@ -113,13 +113,55 @@ public class Stage0Lib{
       new URL[]{ new URL("file:"+file) }, parent
     );
   }
+
+  private static String getVarFromEnv(String envKey) {
+    String value = System.getenv(envKey);
+    if(value==null || value.isEmpty()) {
+      value = System.getenv(envKey.toUpperCase());
+    }
+    return value;
+  }
+
+  private static void setProxyfromPropOrEnv(String envKey, String propKeyH, String propKeyP) {
+    String proxyHost = System.getProperty(propKeyH);
+    String proxyPort = System.getProperty(propKeyP);
+    if((proxyHost==null || proxyHost.isEmpty()) && (proxyPort==null || proxyPort.isEmpty())) {
+      String envVar = getVarFromEnv(envKey);
+      if(envVar != null && !envVar.isEmpty()) {
+        String[] proxy = envVar.replaceFirst("^https?://", "").split(":", 2);
+        System.setProperty(propKeyH, proxy[0]);
+        System.setProperty(propKeyP, proxy[1]);
+      }
+    }
+  }
+
+  public static void installProxySettings() throws URISyntaxException {
+    setProxyfromPropOrEnv("http_proxy", "http.proxyHost", "http.proxyPort");
+    setProxyfromPropOrEnv("https_proxy", "https.proxyHost", "https.proxyPort");
+    String nonHosts = System.getProperty("http.nonProxyHosts");
+    if(nonHosts==null || nonHosts.isEmpty()) {
+      String envVar = getVarFromEnv("no_proxy");
+      if(envVar != null && !envVar.isEmpty()) {
+        System.setProperty("http.nonProxyHosts", envVar.replaceAll(",","|"));
+      }
+    }
+  }
+
+  private static final ProxySelector ps = ProxySelector.getDefault();
+
+  public static HttpURLConnection openConnectionConsideringProxy(URL urlString)
+      throws IOException, URISyntaxException {
+    java.net.Proxy proxy = ps.select(urlString.toURI()).get(0);
+    return (HttpURLConnection) urlString.openConnection(proxy);
+  }
+
   public static void download(URL urlString, Path target, String sha1) throws Exception {
     final Path unverified = Paths.get(target+".unverified");
     if(!Files.exists(target)) {
       new File(target.toString()).getParentFile().mkdirs();
       System.err.println("downloading " + urlString);
       System.err.println("to " + target);
-      final InputStream stream = urlString.openStream();
+      final InputStream stream = openConnectionConsideringProxy(urlString).getInputStream();
       Files.copy(stream, unverified, StandardCopyOption.REPLACE_EXISTING);
       stream.close();
       final String checksum = sha1(Files.readAllBytes(unverified));
--- a/stage1/Stage1Lib.scala
+++ b/stage1/Stage1Lib.scala
@@ -60,7 +60,7 @@ class Stage1Lib( val logger: Logger ) extends BaseLib{
       true
     } else {
       val incomplete = ( target ++ ".incomplete" ).toPath;
-      val connection = url.openConnection.asInstanceOf[HttpURLConnection]
+      val connection = Stage0Lib.openConnectionConsideringProxy(url)
       if(connection.getResponseCode != HttpURLConnection.HTTP_OK){
         logger.resolver(blue("not found: ") ++ url.string)
         false
--- a/stage2/Lib.scala
+++ b/stage2/Lib.scala
@@ -394,7 +394,7 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
     import java.io._
     val url = baseUrl ++ fileName
     System.err.println(blue("uploading ") ++ url.toString)
-    val httpCon = url.openConnection.asInstanceOf[HttpURLConnection]
+    val httpCon = Stage0Lib.openConnectionConsideringProxy(url)
     httpCon.setDoOutput(true)
     httpCon.setRequestMethod("PUT")
     credentials.foreach(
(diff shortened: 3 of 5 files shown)


stage2/Lib.scala
```diff
@@ -394,7 +394,10 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
     import java.io._
     val url = baseUrl ++ fileName
     System.err.println(blue("uploading ") ++ url.toString)
-    val httpCon = url.openConnection.asInstanceOf[HttpURLConnection]
+    // always returns a proxy, in case no proxy is defined, it is the DIRECT proxy object
+    val proxy = ProxySelector.getDefault().select(url.toURI).get(0)
+    //val connection = url.openConnection.asInstanceOf[HttpURLConnection]
```
cvogt: let's not leave lines like this. Git is better for this.

stage1/Stage1Lib.scala
```diff
@@ -60,7 +60,10 @@ class Stage1Lib( val logger: Logger ) extends BaseLib{
       true
     } else {
       val incomplete = ( target ++ ".incomplete" ).toPath;
-      val connection = url.openConnection.asInstanceOf[HttpURLConnection]
+      // always returns a proxy, in case no proxy is defined, it is the DIRECT proxy object
+      val proxy = ProxySelector.getDefault().select(url.toURI).get(0)
+      //val connection = url.openConnection.asInstanceOf[HttpURLConnection]
```
cvogt: same here

nailgun_launcher/Stage0Lib.java
```diff
@@ -113,13 +113,50 @@ public static ClassLoader classLoader( String file, ClassLoader parent ) throws
...
   }
+
+  private static boolean checkEmpty(String value) {
+    return (value==null || value=="");
+  }
+
+  private static void updateProxyPropOrEnv(String envKey) {
+    String base = envKey.replace("_", ".");
+    String pHost = base+"Host";
+    String pPort = base+"Port";
```
cvogt: how about we just use the strings instead of transforming the env var names? Their similarity coincidental.

nailgun_launcher/Stage0Lib.java
```diff
@@ -113,13 +113,50 @@ public static ClassLoader classLoader( String file, ClassLoader parent ) throws
       new URL[]{ new URL("file:"+file) }, parent
     );
   }
+
+  private static boolean checkEmpty(String value) {
```
cvogt: `isEmpty` ?

stage2/Lib.scala
```diff
@@ -394,7 +394,9 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
     import java.io._
     val url = baseUrl ++ fileName
     System.err.println(blue("uploading ") ++ url.toString)
-    val httpCon = url.openConnection.asInstanceOf[HttpURLConnection]
+    // always returns a proxy, in case no proxy is defined, it is the DIRECT proxy object
+    val proxy = ProxySelector.getDefault().select(url.toURI).get(0)
+    val httpCon = url.openConnection(proxy).asInstanceOf[HttpURLConnection]
```
cvogt: how about we move this proxy stuff into a method in Stage0Lib, call it `openConnectionConsideringProxy` and use that everywhere? Better than having 3 places to do this.

nailgun_launcher/Stage0Lib.java
benjaminfrank: wouldn't save me checking for `null` values

cvogt: actually, let's not make this a method. It's really only used in 2 places an re-usability is limited I think.

stage2/Lib.scala
benjaminfrank: Sounds good.

nailgun_launcher/Stage0Lib.java
benjaminfrank: ok

nailgun_launcher/Stage0Lib.java
```diff
@@ -113,13 +113,44 @@ public static ClassLoader classLoader( String file, ClassLoader parent ) throws
...
+        System.setProperty(pHost, s[0]);
+        System.setProperty(pPort, s[1]);
+      }
+    }
+  }
+
+  private static java.net.Proxy proxy = java.net.Proxy.NO_PROXY;
+
+  public static void installProxySettings() throws URISyntaxException {
+    System.err.println("updating proxy settings");
```
cvogt: let's not use `err` for this, otherwise people see it on every startup. I guess since we don't have a logger here, maybe just skip it entirely?

nailgun_launcher/Stage0Lib.java
```diff
@@ -113,13 +113,44 @@ public static ClassLoader classLoader( String file, ClassLoader parent ) throws
...
+
+  private static void updateProxyPropOrEnv(String envKey, String pHost, String pPort) {
+    String value = System.getProperty(pHost);
+    if(value==null || value.isEmpty()) {
+      String proxy = System.getenv(envKey);
+      if(proxy==null || proxy.isEmpty()) {
+        proxy = System.getenv(envKey.toUpperCase());
+      }
+      if(proxy!=null && !proxy.isEmpty()) {
+        String[] s = proxy.replaceFirst("^https?://", "").split(":", 2);
```
cvogt: does this work for http_proxy? looks like a bug

nailgun_launcher/Stage0Lib.java
```diff
@@ -113,13 +113,44 @@ public static ClassLoader classLoader( String file, ClassLoader parent ) throws
...
   }
+
+  private static void updateProxyPropOrEnv(String envKey, String pHost, String pPort) {
+    String value = System.getProperty(pHost);
+    if(value==null || value.isEmpty()) {
+      String proxy = System.getenv(envKey);
+      if(proxy==null || proxy.isEmpty()) {
+        proxy = System.getenv(envKey.toUpperCase());
+      }
+      if(proxy!=null && !proxy.isEmpty()) {
```
cvogt: is this just `else` for the above  `if`

nailgun_launcher/Stage0Lib.java
benjaminfrank: sure its a regex so `s?`means the s is optional

nailgun_launcher/Stage0Lib.java
```diff
@@ -113,13 +113,44 @@ public static ClassLoader classLoader( String file, ClassLoader parent ) throws
       new URL[]{ new URL("file:"+file) }, parent
     );
   }
+
+  private static void updateProxyPropOrEnv(String envKey, String pHost, String pPort) {
```
cvogt: is the name still appropriate? seems like it has nothing to do with proxy anymore

nailgun_launcher/Stage0Lib.java
```diff
@@ -113,13 +113,44 @@ public static ClassLoader classLoader( String file, ClassLoader parent ) throws
...
+
+  public static void installProxySettings() throws URISyntaxException {
+    System.err.println("updating proxy settings");
+    updateProxyPropOrEnv("http_proxy", "http.proxyHost", "http.proxyPort");
+    updateProxyPropOrEnv("https_proxy", "https.proxyHost", "https.proxyPort");
+    // always returns a valid Proxy object
+  }
+
+  public static URLConnection openConnectionConsideringProxy(URL urlString) throws IOException, URISyntaxException {
+    proxy = ProxySelector.getDefault().select(urlString.toURI()).get(0);
```
cvogt: why store this in a global variable?

nailgun_launcher/Stage0Lib.java
benjaminfrank: no, as the value of proxy changes but still can be null or "". What I do here is first check for the lower case version (http_proxy and https_proxy) and in case those env vars are non-existent/empty I try the uppercase variant (HTTP_PROXY and HTTPS_PROXY).
or i didn't get your comment :)

nailgun_launcher/Stage0Lib.java
benjaminfrank: stupid leftover from previous tries, thanks

nailgun_launcher/Stage0Lib.java
cvogt: ah, mutability, right... I see. Happy Java land.

nailgun_launcher/Stage0Lib.java
cvogt: ah, good point, overlooked that. Would anyone end the URL with a slash?

stage2/Lib.scala
```diff
@@ -394,7 +394,7 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
     import java.io._
     val url = baseUrl ++ fileName
     System.err.println(blue("uploading ") ++ url.toString)
-    val httpCon = url.openConnection.asInstanceOf[HttpURLConnection]
+    val httpCon = Stage0Lib.openConnectionConsideringProxy(url).asInstanceOf[HttpURLConnection]
```
cvogt: Alright, one last tiny thing. Mind moving those 2 `.asInstanceOf[HttpURLConnection]` calls into openConnectionConsideringProxy? Then it's ready as far as I am concerned :)!

nailgun_launcher/Stage0Lib.java
benjaminfrank: even is, http is rather forgiving with double slashes, e.g.
http://github.com//foo is the same as http://github.com/foo

benjaminfrank: mildly interesting, that link(s) above are real links and i just made them up when I wrote the comment :)

benjaminfrank:
I redid the PR quite a bit to accomodate for the non_proxy env var.
Other than that, I accounted for your comments/requests and did some real-world testing with a corporate proxy and a setup that requires both an internal and external nexus. works fine for me.

cvogt:
LGTM

PR #204, opened 2016-07-08
AWS Lambda plugin and examples.

mchav:
Requires a file called credentials in the project directory. Everything works fine though.


examples/aws-compile-example/Main.scala
```diff
@@ -0,0 +1,3 @@
+object Main extends App {
+  println("Hello world")
+}
```
cvogt: @mchav posix requires newlines at ends of file :). not a biggie, but may avoid problems later with unix tools.

plugins/aws/README.md
```diff
@@ -0,0 +1,2 @@
+# cbt-aws-example
+Test for cbt cloud compilation.
```
cvogt: could you expand this to describe what steps are needed to try this out?

plugins/aws/RunCbt.java
```diff
@@ -0,0 +1,85 @@
...
+
+public class RunCbt {
+    public static void main(String[] args) {
+        return;
+    }
+
+    @SuppressWarnings("unchecked")
+    public String cbtHandler() {
+        String args = "compile";
+        File dir = new File("."); //to test locally change this to "target"
```
cvogt: Does this work because cbt is run directly on AWS? cwd is that of the nailgun server if one is involved

plugins/aws/RunCbt.java
```diff
@@ -0,0 +1,85 @@
...
+        File[] filesList = dir.listFiles();
+        for (File file : filesList) {
+            File src = file;
+            File dest = new File("/tmp/" + file.getName());
+            try {    
+                copy(src, dest);
+            } catch (Exception e) {
+                e.printStackTrace();
+                return "Copy failed.";
```
cvogt: swallowing an exception and simply printing a string instead hides errors. I would just not catch the exception and let it bubble up to the user instead. Gives more info to help diagnose the issue to the user :).

plugins/aws/RunCbt.java
```diff
@@ -0,0 +1,85 @@
...
+            }
+        }
+
+        File start = new File("/tmp/startCbt.class");
+        String[] splitArgs = args.split("\\s+");
+        Object[] params = { splitArgs };
```
cvogt: at the moment args is hard-coded to "compile", right? What's the plan with this?

plugins/aws/RunCbt.java
```diff
@@ -0,0 +1,85 @@
...
+        for (File file : filesList) {
+            File src = file;
+            File dest = new File("/tmp/" + file.getName());
+            try {    
+                copy(src, dest);
+            } catch (Exception e) {
+                e.printStackTrace();
+                return "Copy failed.";
+            }
+        }
```
cvogt: sounds like this entire thing could be abstracted out as `copyRecursive`.

plugins/aws/RunCbt.java
mchav: @cvogt pass them in as an array or JSON.

plugins/aws/RunCbt.java
```diff
@@ -0,0 +1,85 @@
...
+              
+                int length;
+                //copy the file content in bytes 
+                while ((length = inStream.read(buffer)) > 0){
+              
+                  outStream.write(buffer, 0, length);
+             
+                } 
+                inStream.close();
+                outStream.close();
```
cvogt: can't we use `copy` for this instead of manually handling streams?

cvogt:
great stuff :). Very excited about this!

plugins/aws/RunCbt.java
cvogt: right, but right now this is always Array("compile"), right? because args is hard-coded.

plugins/aws/CodeDeploy.java
```diff
@@ -0,0 +1,108 @@
...
+                        System.out.println(fs.getName());
+                        FileInputStream fis = new FileInputStream(fs);
+                        ZipEntry zipEntry = new ZipEntry(fs.getName());
+                        zos.putNextEntry(zipEntry);
+
+                        byte[] bytes = new byte[1024];
+                        int length;
+                        while ((length = fis.read(bytes)) >= 0) {
+                            zos.write(bytes, 0, length);
+                        }
```
cvogt: this should be replaceable with Files.readAllBytes

plugins/aws/CodeDeploy.java
```diff
@@ -0,0 +1,108 @@
...
+                        zos.closeEntry();
+                        fis.close();
+                    } catch (Exception e) {
+
+                    }
+                }
+                zos.close();
+                fos.close();
+            } catch(IOException ex){
+                ex.printStackTrace();
```
cvogt: Any reason to now just throw the exception :)? I just ended up making every `throws Exception` in my java code and selectively catching exception from the JDK if ever really needed for control flow.

plugins/aws/CodeDeploy.java
```diff
@@ -0,0 +1,108 @@
...
+                        int length;
+                        while ((length = fis.read(bytes)) >= 0) {
+                            zos.write(bytes, 0, length);
+                        }
+
+                        zos.closeEntry();
+                        fis.close();
+                    } catch (Exception e) {
+
+                    }
```
cvogt: catching exceptions and ignoring them is almost always a bad idea in my opinion. hides errors.

plugins/aws/CodeDeploy.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.zip.ZipOutputStream;
+
+import com.amazonaws.AmazonClientException;
+import com.amazonaws.AmazonServiceException;
+import com.amazonaws.auth.profile.ProfileCredentialsProvider;
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.AmazonS3Client;
+import com.amazonaws.services.s3.model.PutObjectRequest;
+
+public class CodeDeploy {
```
cvogt: Any reason to use Java, not Scala for this?

plugins/aws/CodeDeploy.java
```diff
@@ -0,0 +1,108 @@
...
+         } catch (AmazonServiceException ase) {
+            System.out.println("Caught an AmazonServiceException, which " +
+                    "means your request made it " +
+                    "to Amazon S3, but was rejected with an error response" +
+                    " for some reason.");
+            System.out.println("Error Message:    " + ase.getMessage());
+            System.out.println("HTTP Status Code: " + ase.getStatusCode());
+            System.out.println("AWS Error Code:   " + ase.getErrorCode());
+            System.out.println("Error Type:       " + ase.getErrorType());
+            System.out.println("Request ID:       " + ase.getRequestId());
```
cvogt: System.err is probably more appropriate here

plugins/aws/build/build.scala
```diff
@@ -0,0 +1,24 @@
...
+
+class Build( context: Context ) extends BasicBuild( context ) with Plugin {
+  override def runClass = "Deploy" // make runCbt to test locally
+  override def dependencies = (
+    super.dependencies // don't forget super.dependencies here
+    ++
+    Resolver( mavenCentral ).bind(
+      MavenDependency( "com.amazonaws", "aws-java-sdk", "1.11.8" ),
+      MavenDependency( "com.amazonaws", "aws-lambda-java-core", "1.1.0"),
+      MavenDependency( "com.amazonaws", "aws-java-sdk-s3", "1.11.9"),
```
cvogt: doesn't this automatically pull in the sdk? Also the versions should match :)

plugins/aws/aws.scala
```diff
@@ -0,0 +1,322 @@
...
+import com.amazonaws.regions.Region
+import com.amazonaws.regions.Regions
+import com.amazonaws.services.lambda._
+import com.amazonaws.services.lambda.model._
+import com.amazonaws.services.s3.AmazonS3
+import com.amazonaws.services.s3.AmazonS3Client
+import com.amazonaws.services.s3.model._
+
+
+abstract case class OnAwsLambda( targetBuild: BaseBuild )(implicit logger: Logger) extends DependencyImplementation {
```
cvogt: why is this abstract?

PR #208, opened 2016-07-18
Remove cbt command from argument.

mchav:


stage2/BasicBuild.scala
```diff
@@ -134,7 +134,7 @@ trait BaseBuild extends DependencyImplementation with BuildInterface with Trigge
   }
 
   def runClass: String = "Main"
-  def run: ExitCode = lib.runMainIfFound( runClass, context.args, classLoader(context.classLoaderCache) )
+  def run: ExitCode = lib.runMainIfFound( runClass, context.args.drop(1), classLoader(context.classLoaderCache) )
```
cvogt: That's not the right place to fix it. This would mean everything else still sees that task as part of the args. The right place is Stage2.scala. I have a fix for it in a branch, but it's as minor as the one above, so feel free, if you want. Closing this one for now.

PR #128, opened 2016-05-22
Bash completions.

mchav:
Bash completions for all commands in usage. Subcommands not yet supported.

PR #211, opened 2016-08-14
Completions

cvogt:
supersedes #128


README.md
```diff
@@ -176,3 +176,19 @@ solution could be code generating traits at build-time and mixing them
...
 class loading is not rocket science. But there may be simpler solutions
 for the cases at hand. And they are edge cases anyways.
+
+Bash completions
+----------------
+To auto-complete cbt task names in bash do this:
+
+```
+mkdir ~/.bash_completion.d/
+cp shell-integration/cbt-completions.bash ~/.bash_completion.d/
```
mchav: This depends on the the OS, doesn't it? Mine is in `/etc`

README.md
```diff
@@ -176,3 +176,19 @@ solution could be code generating traits at build-time and mixing them
...
+```
+
+Add this to you .bashrc
+```
+for f in ~/.bash_completion.d/*; do
```
mchav: And once it's in `etc` there's no need to do this.

cvogt: right, this is a user specific one in addition to the global one (which seems to be located in different places dependent on the operating system).

The way described here definitely works, does not require sudo and is independent of the OS. People can totally put it into a global place instead, but I think that will be clear to any more advanced bash users.

PR #214, opened 2016-08-15
private git repos

cvogt:


cvogt:
supersedes #127

mchav:
This one too.

PR #224, opened 2016-09-14
WartRemover plugin implementation

melrief:
A first implementation of the WartRemover plugin. Things that need to be addressed to get a fully working plugin:

1. I'm currently using the `WartTraverser` type instead of `Wart` which means that many utilities like `Warts.all` are still missing
2. Testing all the options


cvogt:
awesome stuff @melrief :). Closing in favor of #225, which adds a minor commit to fixes the test failure.

PR #241, opened 2016-09-28
Avoid getting started confusion like …

cvogt:
https://gitter.im/cvogt/cbt?at=57ec271754302caa6f2950b7

### DIFF
--- a/stage2/Scaffold.scala
+++ b/stage2/Scaffold.scala
@@ -34,16 +34,16 @@ class Build(val context: Context) extends BaseBuild{
     super.dependencies ++ // don't forget super.dependencies here
     Seq(    
       // source dependency
-      DirectoryDependency( projectDirectory ++ "/subProject" )
+      // DirectoryDependency( projectDirectory ++ "/subProject" )
     ) ++
     Resolver( mavenCentral ).bind(
       // CBT-style Scala dependencies
-      ScalaDependency( "com.lihaoyi", "ammonite-ops", "0.5.5" )
-      MavenDependency( "com.lihaoyi", "ammonite-ops_2.11", "0.5.5" )
+      // ScalaDependency( "com.lihaoyi", "ammonite-ops", "0.5.5" )
+      // MavenDependency( "com.lihaoyi", "ammonite-ops_2.11", "0.5.5" )
 
       // SBT-style dependencies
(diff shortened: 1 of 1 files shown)


stage2/Scaffold.scala
```diff
@@ -34,16 +34,16 @@ class Build(val context: Context) extends BaseBuild{
     super.dependencies ++ // don't forget super.dependencies here
     Seq(    
       // source dependency
-      DirectoryDependency( projectDirectory ++ "/subProject" )
+      // DirectoryDependency( projectDirectory ++ "/subProject" )
     ) ++
     Resolver( mavenCentral ).bind(
```
mariussoutier: I'd suggest adding a comment here, I think the `Resolver.bind` is a very nice feature in contrast to sbt. I assume it means you can control directly how to resolve your deps, so in a company I'd use Artifactory and cbt wouldn't try to load over the Internet?

stage2/Scaffold.scala
```diff
@@ -34,16 +34,16 @@ class Build(val context: Context) extends BaseBuild{
...
+      // ScalaDependency( "com.lihaoyi", "ammonite-ops", "0.5.5" )
+      // MavenDependency( "com.lihaoyi", "ammonite-ops_2.11", "0.5.5" )
 
       // SBT-style dependencies
-      "com.lihaoyi" %% "ammonite-ops" % "0.5.5"
-      "com.lihaoyi" % "ammonite-ops_2.11" % "0.5.5"
+      // "com.lihaoyi" %% "ammonite-ops" % "0.5.5"
+      // "com.lihaoyi" % "ammonite-ops_2.11" % "0.5.5"
     )
   */
```
mariussoutier: I think the /* .. */ could be removed now.

stage2/Scaffold.scala
cvogt: Yep indeed. There is no global list of resolvers, instead you bind individual dependencies (and their transitive dependencies) to individual resolvers.

stage2/Scaffold.scala
cvogt: done https://github.com/cvogt/cbt/pull/244

stage2/Scaffold.scala
cvogt: added a comment https://github.com/cvogt/cbt/pull/244

PR #247, opened 2016-10-01
Sonatype release plugin

### LINKED ISSUE #10: sonatype support library
https://github.com/xerial/sbt-sonatype

rockjam:
This PR contains implementation of Sonatype plugin, that allows to pulish artifacts to Sonatype repo and release it to Maven Central repository.

There are two commands:
`cbt sonatypePublishSigned` - to publish signed artifacts to staging repo
`cbt sonatypeRelease` - to close staging repo, sync artifacts to Maven Central and drop this temporary staging repo.

To show debug log, use -Dlog=Sonatype flag.
Sonatype credentials should be in `~/.sonatype/.credentials` file, or could be provided via `SONATYPE_USERNAME` and `SONATYPE_PASSWORD` env variables.

Fixes #10


stage2/Lib.scala
```diff
@@ -450,12 +449,12 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
...
 
   def uploadAll(url: URL, nameAndContents: Seq[(String, Array[Byte])], credentials: Option[String] = None ): Unit =
-    nameAndContents.map{ case(name, content) => upload(name, content, url, credentials ) }
+    nameAndContents.foreach { case (name, content) => upload(name, content, url, credentials ) }
 
   def upload(fileName: String, fileContents: Array[Byte], baseUrl: URL, credentials: Option[String] = None): Unit = {
     import java.net._
     import java.io._
-    val url = baseUrl ++ fileName
+    val url = baseUrl ++ "/" ++ fileName
```
cvogt: does this work? I think it needs `++ ("/" ++ fileName)`

stage2/Publish.scala
```diff
@@ -35,37 +37,10 @@ trait Publish extends PackageJars{
...
-  }
-
-  def publishLocal: Unit = {
+  def publishLocal: Unit =
     lib.publishLocal( sourceFiles, `package` :+ pom, context.paths.mavenCache, releaseFolder )
-  }
-
-  def publishSnapshotLocal: Unit = {
-    copy( context.copy(version = Some(version+"-SNAPSHOT")) ).publishLocal
-  }
```
cvogt: was this suppose to go away?

plugins/sonatype-release/SonatypeRelease.scala
```diff
@@ -0,0 +1,438 @@
...
+  def profileName: String = groupId
+
+  // sonatype service uri
+  def sonatypeRepoUri: String = "https://oss.sonatype.org/service/local"
+
+  /**
+    * login:password pair for Sonatype access.
+    * Order of credentials lookup:
+    * • environment variables SONATYPE_USERNAME and SONATYPE_PASSWORD
+    * • ~/.sonatype/.credentials file
```
cvogt: what if people want different credentials for different repositories?

stage2/Lib.scala
rockjam: Yep, it works, `baseUrl ++ "/"` is of type java.net.URL and we get ++ via extension class

stage2/Publish.scala
rockjam: Not because of Sonatype plugin actually. My usual way to publish local snapshot is to change version in build file and publish it. I usually don't keep non-snapshot versions locally. When you need to publish real version, you remove snapshot, publish real version, increment version and add -SNAPSHOT again. It's more like dev version you can publish to local repo at any time

plugins/sonatype-release/SonatypeRelease.scala
```diff
@@ -0,0 +1,438 @@
...
+        ),
+        credentials = Some(sonatypeCredentials)
+      )
+      System.err.println(lib.green("Successfully published artifacts to Sonatype."))
+      ExitCode.Success
+    } else {
+      System.err.println(lib.red("Sources are empty, won't publish empty jar."))
+      ExitCode.Failure
+    }
+  }
```
cvogt: can we move these two methods into SonatypeLib and just have small forwarder methods with the same name here calling the real ones with whatever arguments they require? I'd like to keep the mixing part of cbt plugins as slim as possible, really just for configuration, not logic. This way stuff is less coupled to builds and OO and could be at some point used differently.

plugins/sonatype-release/SonatypeRelease.scala
```diff
@@ -0,0 +1,438 @@
...
+          val newRetries = retriesLeft - 1
+          log(s"Failed with exception: $e, will retry $newRetries times; waiting: $delay")
+          Thread.sleep(delay)
+
+          withRetry(newDelay, newRetries)(f)
+        }
+    }
+  }
+
+}
```
cvogt: how about we move everything except for `trait SonatypeRelease` into a package `cbt.sonatype` to avoid blowing up cbt's namespace?

plugins/sonatype-release/SonatypeRelease.scala
rockjam: That's tricky question. This should be handled somehow, but I'm sure that people shouldn't put .credentials file under version control in project directory

stage2/Lib.scala
```diff
@@ -430,11 +431,9 @@ final class Lib(logger: Logger) extends Stage1Lib(logger) with Scaffold{
     }
   }
 
-  def publishSigned( sourceFiles: Seq[File], artifacts: Seq[File], url: URL, credentials: Option[String] = None ): Unit = {
+  def publishSigned( artifacts: Seq[File], url: URL, credentials: Option[String] = None ): Unit = {
     // TODO: make concurrency configurable here
-    if(sourceFiles.nonEmpty){
-      publish( artifacts ++ artifacts.map(sign), url, credentials )
-    }
+    publish( artifacts ++ artifacts.map(sign), url, credentials )
```
cvogt: this check was here, so we don't publish empty jars, when there are not source files. Any particular reason you removed it?

plugins/sonatype-release/SonatypeRelease.scala
rockjam: Do you mean `sonatypePublishSigned` and `sonatypeRelease`?

stage2/Lib.scala
cvogt: This worked before, didn't it? Why is the "/" needed now?

plugins/sonatype-release/SonatypeRelease.scala
rockjam: :+1:

stage2/PackageJars.scala
```diff
@@ -1,8 +1,10 @@
 package cbt
 import java.io.File
+
 // would love to call this just `Package` but that conflicts with scala package objects.
 trait PackageJars extends BaseBuild with ArtifactInfo{
   def name: String
+  // TODO: why not final?
```
cvogt: what not final? `artifactid`? People may want it to be different from `name`, which is used in multiple places, e.g. the pom file.

stage2/Lib.scala
rockjam: One reason - we pass `sourceFiles ` **just** to check if those are empty, it makes me think that we will use `sourceFiles` somehow, but we throw them away right there. Can't explain better)

stage2/Lib.scala
rockjam: Convention, to never put / at the end of uri. If people agree not to put / at the end of uri working with it becomes more predictable.

```
val google = "https://google.com"
val doodles = google + "/doodles"
def findDoodle(name:String) = doodles + "?name=new_year_2016"
```

stage2/PackageJars.scala
rockjam: Yes, just wanted to clarify this. Thanks, will remove it

stage2/Publish.scala
cvogt: right, I added this so you don't have to change the build file :). I prefer typing `cbt publishSnapshotLocal` over having to change the build file. Mind undoing the remove :)?

stage2/Lib.scala
cvogt: Ah, now I understand. Doesn't it make more sense that the caller of this function has to provide the url including a trailing slash, because it is a directory into which the files should go?

cvogt: or do you think that's error prone

plugins/sonatype-release/SonatypeRelease.scala
cvogt: sonatypeCredentials and sonatypeRepoUri as well, if someone wants to re-use this without the trait he needs it all. But we should still put proxy methods into the trait so people can override them, but the implementations / values should come from SonatypeLib, WDYT?

stage2/Lib.scala
cvogt: yeah, I wasn't happy with this either. Does your code prevent uploads of empty jars in a different way or does it not prevent it?

stage2/Lib.scala
cvogt: I don't have a strong opinion on this

stage2/Publish.scala
rockjam: Sure

plugins/sonatype-release/SonatypeRelease.scala
rockjam: `~/.cbt/.sonatype/.credentials` maybe? Put all sonatype related stuff to single directory. Maybe we could provide multiple credentials in this directory or something else in future.

rockjam: not sure about .sonatype or just sonatype in path

plugins/sonatype-release/SonatypeRelease.scala
rockjam: Sure, the logic should go to lib

stage2/Lib.scala
rockjam: Well, I just check that sources aren't empty at call site. We can leave old logic here, and have duplication in plugin's code

stage2/Publish.scala
```diff
@@ -35,37 +37,10 @@ trait Publish extends PackageJars{
...
-
-  protected def sonatypeCredentials: Option[String] = {
-    // FIXME: this should probably not use cbtHome, but some reference to the system's host cbt
-    Some(new String(readAllBytes((context.cbtRootHome ++ "/sonatype.login").toPath)).trim)
-  }
+  private val releaseFolder = s"/${groupId.replace(".","/")}/${artifactId}_$scalaMajorVersion/$version/"
 
-  def publishSnapshot: Unit = {
-    copy( context.copy(version = Some(version+"-SNAPSHOT")) ).publishUnsigned
-  }
```
rockjam: Need to check that. I believe that SNAPSHOT publish happens in same way as regular one.

examples/sonatype-release-example/build/build.scala
```diff
@@ -0,0 +1,31 @@
...
+      "GMT+3",
+      new URL("https://github.com/rockjam")
+    )
+  )
+  def inceptionYear: Int = 2016
+  def licenses: Seq[cbt.License] = Seq(License.Apache2)
+  def organization: Option[cbt.Organization] = None
+  def scmConnection: String = ""
+  def scmUrl: String = "https://github.com/rockjam/cbt-sonatype.git"
+  def url: java.net.URL = new URL("https://github.com/rockjam/cbt-sonatype")
```
rockjam: Seen it, just wanted to show example with full settings required for publishing. Can put a hint for users to use `GithubPom`

plugins/sonatype-release/SonatypeRelease.scala
cvogt: Actually I guess one part starting with a dot in the path is enough. I thought about `~/.cbt/sonatype/credentials` before, but then I thought a top-level file is totally fine and a sonatype subfolder is just adding slight complexity where it is not needed at the moment. A top-level file in `~/.cbt/` is easier to find and manage I think. But it doesn't make a big difference. My favorite would be `~/.cbt/sonatype-credential` now, but I am ok with other choices inside of `~/.cbt/` as well.

stage2/Lib.scala
cvogt: checking at the call site is fine with me, no need to duplicate

stage2/Publish.scala
cvogt: I think it goes to a different URL

examples/sonatype-release-example/build/build.scala
cvogt: I see. Cool. Yeah a comment is probably helpful :)

rockjam:
@cvogt updated

plugins/sonatype-release/src/cbt/sonatype/SonatypeLib.scala
```diff
@@ -0,0 +1,151 @@
...
+
+        Try(toRelease.get) match {
+          case Success(repo) =>
+            sonatypeApi.finishRelease(repo, profile)
+            System.err.println(lib.green(s"Successfully released ${groupId}/${artifactId} v:${version}"))
+            ExitCode.Success
+          case Failure(_) =>
+            System.err.println(lib.red("Wrong repository number, try again please."))
+            ExitCode.Failure
+        }
```
cvogt: using .map(...).getOrElse(...) seems more appropriate here than catching the None.get get exception, wrapping it in a try and matching on that

rockjam: Sure, it will be better. That was adaptation of previous code, I didn't realise other solution at the moment

cvogt: Yep, 4 eyes see more than 2 :). That's the beauty of code reviews.

PR #253, opened 2016-10-02
Run

cvogt:
@rockjam wanna give this quick review :)? pickOne may be useful to you.

### DIFF
--- a/stage1/Stage1Lib.scala
+++ b/stage1/Stage1Lib.scala
@@ -118,6 +118,43 @@ class Stage1Lib( val logger: Logger ) extends BaseLib{
     }
   }
 
+  def mainClasses(compileTarget : File, classLoader : ClassLoader): Seq[String] = {
+    val targetLength = compileTarget.toString.length + 1 // compile target and slash
+
+    def className(file : File) : String = {
+      file.toString.dropRight(".class".length).drop(targetLength).replace(File.separator, ".")
+    }
+
+    /* for packaged applications */
+    def retrieveFiles(file: File): Seq[File] = {
+      file.listFiles.toSeq.map(currFile => {
+          if (currFile.isDirectory) retrieveFiles(currFile)
+          else Seq(currFile)
+        }
+      ).flatten
+    }
+
+    def classesWithMain(target : File, classLoader : ClassLoader): Seq[File]  = {
+      def hasMainClass(file : File) : Boolean = {
+        val name = className(file)
+        val inspectClass = classLoader.loadClass(name)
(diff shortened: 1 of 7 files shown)


rockjam:
@cvogt will take a look

stage1/Stage1Lib.scala
```diff
@@ -118,6 +112,67 @@ class Stage1Lib( val logger: Logger ) extends BaseLib{
     }
   }
 
+  /** shows an interactive dialogue in the shell asking the user to pick one of many choices */
+  def pickOne[T]( msg: String, choices: Seq[T] )( show: T => String ): Option[T] = {
+    if(choices.size == 0) None else if(choices.size == 1) Some(choices.head) else {
+      Option(System.console).map{
```
rockjam: Why System.console and not scala.Console?

stage1/Stage1Lib.scala
```diff
@@ -118,6 +112,67 @@ class Stage1Lib( val logger: Logger ) extends BaseLib{
...
+  /** interactively pick one main class */
+  def runClass( mainClasses: Seq[Class[_]] ): Option[Class[_]] = {
+    pickOne( "Which one do you want to run?", mainClasses )( _.toString )
+  }
+
+  def mainClasses( targetDirectory: File, classLoader : ClassLoader ): Seq[Class[_]] = {
+    val arrayClass = classOf[Array[String]]
+    val unitClass = classOf[Unit]
+
+    listFilesRecursive(targetDirectory)
```
rockjam: Indent on .filter

test/test.scala
```diff
@@ -1,4 +1,5 @@
-import cbt._
+package cbt
+package test
```
rockjam: Which style is encouraged in cbt codebase? 
`package cbt.some.thing` 
or 
```
package cbt
package some
package thing
```

stage1/Stage1Lib.scala
cvogt: The io stuff in scala's standard library feels mostly ad-hoc and outdated, so I tend to use the JDK ones.

test/test.scala
cvogt: @rockjam it's not a style. It is semantically different.
```
package cbt
package test
```
is equivalent to 
```
package cbt.test
import cbt._
```

So my view is let's only it when we need the import equivalent

stage1/Stage1Lib.scala
cvogt: fixed

test/test.scala
rockjam: Well, that was a discovery for me!

PR #255, opened 2016-10-02
upgrade scalafmt 0.3.1 -> 0.4.2; load scalafmt config from config file

rockjam:
Did some improvements to scalafmt plugin. Now, it uses config from .scalafmt.conf by default, as it does sbt-plugin and IDEA plugin.


rockjam:
It also adds `scalafmt-cli` dependency. I couldn't avoid it, because of `org.scalafmt.cli.StyleCache` is there.

cvogt:
Nice! cc @olafurpg

PR #265, opened 2016-10-04
repl task in BaseBuild

rockjam:
Added task for project-aware REPL. Uses project's classpath and project's scala version

### DIFF
--- a/stage2/BasicBuild.scala
+++ b/stage2/BasicBuild.scala
@@ -2,12 +2,6 @@ package cbt
 
 import java.io._
 import java.net._
-import java.nio.file.{Path =>_,_}
-import java.nio.file.Files.readAllBytes
-import java.security.MessageDigest
-import java.util.jar._
-
-import scala.util._
 
 class BasicBuild(val context: Context) extends BaseBuild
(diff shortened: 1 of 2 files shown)


stage2/BasicBuild.scala
```diff
@@ -156,7 +150,16 @@ trait BaseBuild extends BuildInterface with DependencyImplementation with Trigge
     )
   }
 
-  def test: Option[ExitCode] = 
+  def repl: ExitCode = {
+    lib.consoleOrFail("Use `cbt direct repl` instead")
+    val scalac = new ScalaCompilerDependency(context.cbtHasChanged, context.paths.mavenCache, scalaVersion)
+    val _args = Seq("-cp", s"${scalac.classpath.string}:${classpath.string}") ++ context.args
```
cvogt: I believe this should be

```
val _args = Seq("-bootclasspath", s"${scalac.classpath.string}","-classpath", s"${classpath.string}") ++ context.args
```

Otherwise scalac ends up on the repl classpath even if the lib doesn't require it. 

(also let's use the long option names, which are optimized for readability rather than writability IMHO)

rockjam: Ah, sure, forgot about bootclasspath

Issue #172, opened 2016-06-23
tasks that take arguments, callable from command line


cvogt:
automatic generation of cli interface for tasks that take arguments, e.g.

```
def compile(target: File) = ...
```
```
cbt compile --target=/path/to/target
```

also see http://www.lihaoyi.com/Ammonite/#ScriptArguments

cvogt:
short names could be determined automatically

```
cbt compile -t /path/to/target
```

alternatively, the command line syntax could just use scala syntax
```
cbt compile("/path/to/target")
cbt compile(target="/path/to/target")
```

cvogt:
May be as simple as integrating this:
https://github.com/lihaoyi/Ammonite/blob/test/repl/src/main/scala/ammonite/repl/Router.scala#L1 
![image](https://cloud.githubusercontent.com/assets/274947/16522305/04910672-3f6b-11e6-82ba-f69686b6cb58.png)
![image](https://cloud.githubusercontent.com/assets/274947/16522311/0fffc516-3f6b-11e6-91c4-eb2fdc94d1e8.png)
source: https://gitter.im/cvogt/cbt?at=57762f29632b75030fc8ad34

rockjam:
Currently, workaround to use parameters of some kind is to use environment variables and access them via `System.getenv` . It doesn't look nice, but can be used in "emergency"

in `build.scala`:
```
  def foo: Int = {
    val optRes = for {
      a <- Option(System.getenv("CBT_A"))
      b <- Option(System.getenv("CBT_B"))
    } yield a.toInt + b.toInt
    optRes.getOrElse(throw new Exception("Pass env varialbes CBT_A and CBT_B to get result"))
  }
```
Call task like this: `CBT_A=2 CBT_B=3 cbt foo`

Of course, it's not a solution, more and ugly hack

rockjam:
Ah, this appear to work without nailgun, I had some difficulties to make it work with nailgun

cvogt:
to access env vars through nailgun you need to access ngcontext instead of the jdk api. Also see https://github.com/cvogt/cbt/issues/197 and https://github.com/cvogt/cbt/projects

But you can also just pass args to CBT via the command line. In your build they are accessible via `context.args`. CBT also passes them further to main methods run via `cbt run`.

rockjam:
Oh, I forgot about `context.args` . Seen them before

cvogt:
vaguely related to https://github.com/cvogt/cbt/issues/210

PR #266, opened 2016-10-04
System properties

rockjam:
This one is work in progress. I'm not sure if it is right way to pass system properties to build tasks and user's code. In https://github.com/rockjam/cbt/blob/system-properties/examples/system-properties/README.md I described how it works in current implementation.

### DIFF
--- a/stage2/Stage2.scala
+++ b/stage2/Stage2.scala
@@ -1,5 +1,6 @@
 package cbt
-import java.io._
+
+import scala.collection.JavaConverters._
 
 object Stage2 extends Stage2Base{
   def getBuild(__context: java.lang.Object, _cbtChanged: java.lang.Boolean) = {
@@ -32,6 +33,7 @@ object Stage2 extends Stage2Base{
     val context: Context = ContextImplementation(
       args.cwd,
       args.cwd,
+      args.props.asJava,
       args.args.drop( taskIndex +1 ).toArray,
       logger.enabledLoggers.toArray,
       logger.start,
@@ -46,6 +48,10 @@ object Stage2 extends Stage2Base{
       args.compatibilityTarget,
       null
     )
+
+    // set System.properties passed as argument to task.
+    context.propsMap().asScala foreach { case (k, v) => System.setProperty(k, v) }
+
     val first = lib.loadRoot( context )
     val build = first.finalBuild
 
@@ -55,7 +61,7 @@ object Stage2 extends Stage2Base{
           v => new lib.ReflectBuild(
             build.copy(context.copy(scalaVersion = Some(v)))
           ).callNullary(task)
-        }.filter(_ != ExitCode.Success).headOption getOrElse ExitCode.Success
+        }.find(_ != ExitCode.Success) getOrElse ExitCode.Success
       } else {
         new lib.ReflectBuild(build).callNullary(task)
       }
@@ -65,7 +71,7 @@ object Stage2 extends Stage2Base{
       if (loop) {
         // TODO: this should allow looping over task specific files, like test files as well
         val triggerFiles = first.triggerLoopFiles.map(lib.realpath)
-        val triggerCbtFiles = Seq( nailgun, stage1, stage2 ).map(lib.realpath _)
+        val triggerCbtFiles = Seq( nailgun, stage1, stage2 ).map(lib.realpath)
         val allTriggerFiles = triggerFiles ++ triggerCbtFiles
 
         logger.loop("Looping change detection over:\n - "++allTriggerFiles.mkString("\n - "))
--- a/compatibility/Context.java
+++ b/compatibility/Context.java
@@ -1,11 +1,13 @@
 package cbt;
 import java.io.*;
+import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
 // TODO: try to reduce the number of members
 public abstract class Context{
   public abstract File projectDirectory();
   public abstract File cwd();
+  public abstract Map<String, String> propsMap();
   public abstract String[] argsArray();
   public abstract String[] enabledLoggersArray();
   public abstract Long startCompat();
--- a/examples/system-properties/Main.scala
+++ b/examples/system-properties/Main.scala
@@ -1,0 +1,12 @@
+
+object Main extends App {
+  val foo: Option[String] = Option(System.getProperty("foo"))
+  val bar: Option[String] = Option(System.getProperty("bar"))
+
+  println(s"=== Custom properties, foo: ${foo}, bar: ${bar}")
+
+  val timezone: String = System.getProperty("user.timezone")
+  val os: String = System.getProperty("os.name")
+
+  println(s"=== System properties, timezone: ${timezone}, os: ${os}")
+}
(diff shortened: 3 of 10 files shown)


stage2/Stage2.scala
```diff
@@ -46,6 +48,10 @@ object Stage2 extends Stage2Base{
       args.compatibilityTarget,
       null
     )
+
+    // set System.properties passed as argument to task.
+    context.propsMap().asScala foreach { case (k, v) => System.setProperty(k, v) }
```
rockjam: I really don't know where is the best place for it

cvogt:
Very interesting.

This PR seems to achieve the same thing for the nailgun client: https://github.com/martylamb/nailgun/pull/17 but it is not merged. Possibly because system properties are global mutable stage and multiple clients would interfere with each other.

We have the same problem with CBT, so it is not advisable to use system properties with CBT. At the same time we should have a mechanism to provide them for things that already rely on them.

I wonder if there is some way to highjack the system property mechanism in the jvm to make reject mistakes (e.g. reject thread writing or reading system properties other still active threads have set) or fix the behavior by scoping them to threads via thread locals and possibly copying the local scope when a thread creates a new thread. (Only slightly related: we might have to consider similar crazy hackery to make System.console work).

I am supportive with this PR, I need more time to review it :). Good stuff @rockjam

cvogt:
The before-mentioned high-jacking may involve using the SecurityManager to detect system property reads and writes and/or byte-code rewriting to highjack the mechanism.

cvogt:
@rockjam FYI, this PR fails because the change to ContextImplementation broke source compatibility. The way to sent another commit on top of this that replaces references to old versions with the previous commit of this PR. In particular search for the two occurences of `1f4f6097d3ca682d6fd20a7cc6dd277832350827` and replace it with `f57386c2ba95bf47309f930eeaa48fd4d88368ab`. But let's delay this until review is done, because if it involves rebasing the commit hashes will change.

cvogt:
@rockjam sorry for the delay. So I have thought about this. One thing I don't like about this yet is that system properties are global mutable state, which permanently affects the background process when people use nailgun. I think people should ideally be conscious that they are affecting global mutable state shared by all cbt instances running, but a persistent background process is unusual for many people.

I know you are using cbt without nailgun, so you are not affected by this, but system properties don't work well with nailgun for that reason. So how about we do this:

We separate the mechanism to affect system properties from the mechanism we provide properties to cbt. We have two mechanisms. CBT's own mechanism is not global but scoped to the cbt instance and disappears when it terminates eventhough the nailgun background process keeps running. We also provide a separate mechanism to affect system properties but urge people to only use it when they have to affect system properties out of their control and warn them about the implications.

In detail this could look like this:

`cbt -Xlog=all -Dscala.color=true repl`

`-X` are the cbt specific properties that are NOT turned into system properties (the name is borrowed from the java/scala equivalents). `-Xlog` sets the loggers, just like`-Dlog` currently, independent for every cbt run, no global mutable state. `-Dscala.color=true` will actually set a system property. However cbt exits with an error when trying to use `-D` when running on nailgun. So you can't use it. Instead you'll have to do this:

`cbt setProperties -Dscala.color=true` which returns an error when NOT running on nailgun, but sets the system property otherwise.

This highlights very well that system properties are performing mutation.

If we want a way to have the exact same cbt command line invocation work with and without nailgun, we can consider an additional flag, e.g. `-Xunsafe` to raise awareness of the risk.

`cbt -Xlog=all -Xunsafe -Dscala.color=true repl`

This would then work regardless of nailgun or not and set the system properties.

Also, we should probably only allow `-X` and `-D` to be provided BEFORE task name, so 

`cbt <allowed here> <task name> <but not here>`

so that the task could itself take -X or -D arguments, e.g. if it is the Scala repl, etc.

WDYT?

cvogt:
now I am thinking maybe we should remove the `X` from regular cbt arguments, so just `-log=all` and `-unsafe`. And we should not pass things like the value of `log` through in an untyped map. 
But I can see having `-X...` args in an untyped map for user extensibility.

rockjam:
@cvogt
System you described seems difficult to me. There are some things I agree, but want to make it simpler somehow, and I have some suggestions.

1. I agree to separate cbt's properties from system properties, so we could do:

  ```
  cbt -log=all -debug=true repl -Dscala.color=true
  ```
  where `-log` and `-debug` flags are cbt's, and `-Dscala.color=true` is global flag passed to JVM

2. Don't let people use system properties with nailgun at all.
  We already have warnings in cbt, saying: "use cbt direct instead". We could do the same in cases when user need to pass system properties.
  So, allow only:
  ```
  cbt direct repl -Dscala.color=true
  ```
  but not:
  ```
  cbt repl -Dscala.color=true
  ```
3. Don't let people to pass system properties to task itself.
  > Also, we should probably only allow -X and -D to be provided BEFORE task name, so
  > ```cbt <allowed here> <task name> <but not here>```
  > so that the task could itself take -X or -D arguments, e.g. if it is the Scala repl, etc.
  
  Want to find out, if we are on the same page. AFAIK, we just pass system properties globally, to instance of JVM that is running current task. System properties will be the same inside cbt's task and inside the thing that task may run. For example, when we execute `cbt run -Duser.timezone=UTC`, we'll get same value of `System.getProperty("user.timezone")` inside `run` task and inside `Main` class. This is how it works.
  
  I also think that we could avoid passing system properties to tasks in favour args(#172) of environment variables.

cvogt:
> 1.

Sounds good!

> Don't let people use system properties with nailgun at all.

Good idea. Let's start with that. We can always start supporting it in the future if there is need, but let's keep it simple for now. 👍 

> I also think that we could avoid passing system properties to tasks in favour args(#172) of environment variables.

I think we need to support it for interaction with existing libraries like logback.

> 3.

I just tried this:

```
$ echo 'println(System.getProperty("foo") + args)' > test.scala
$ scala -Dfoo="hello" test.scala asdf
hello WrappedArray(asdf)
$ scala test.scala asdf -Dfoo="hello"
hello WrappedArray(asdf, -Dfoo=hello)
```

So interestingly Scala turns -D into system properties AND passes them as arguments to the application. Java handles it differently:

```
$ echo 'object Foo extends App{ println(System.getProperty("foo") + " " + args.toSeq) }' > test.scala
$ scalac test.scala asdf
$ java -cp (cbt tools resolve org.scala-lang:scala-library:2.10.6):. -Dfoo="hello" Foo asdf
hello WrappedArray(asdf)
$ java -cp (cbt tools resolve org.scala-lang:scala-library:2.10.6):. Foo asdf -Dfoo="hello"
null WrappedArray(asdf, -Dfoo=hello)
```

So Java only interprets them as system properties when passed BEFORE the name of the application.

I feel `java` acts more reasonably here. With `scala` as seen above there is no way to pass `-Dfoo=hello` to a program without making it a system property. An application (or task in cbt land) may however require a `-D` argument for some reason, e.g. because it passes arguments to a non-java command line tool like `ssh`, where `ssh -D...` has a different meaning. `cbt` currently prevents this. `scala` seems to entangle it with system properties. `java` seems to be doing the right thing and we should probably make `cbt` do the same thing.

cvogt:
moving to milestone 1.0, since this would clash too much with the planned refactorings afterwards

PR #271, opened 2016-10-07
IOException handling in Stage1Lib.download

tim-zh:
See [failed build](https://circleci.com/gh/cvogt/cbt/489).

Should there be a retry attempt?

### DIFF
--- a/stage1/Stage1Lib.scala
+++ b/stage1/Stage1Lib.scala
@@ -69,10 +69,14 @@ class Stage1Lib( val logger: Logger ) extends BaseLib{
         logger.resolver(blue("to ") ++ target.string)
         target.getParentFile.mkdirs
         val stream = connection.getInputStream
-        try{
-          Files.copy(stream, incomplete, StandardCopyOption.REPLACE_EXISTING)
-        } finally {
-          stream.close()
+        try {
+          try
+            Files.copy(stream, incomplete, StandardCopyOption.REPLACE_EXISTING)
+          finally
+            stream.close()
+        } catch {
+          case e: IOException =>
+            System.err.println(red("downloading failed ") ++ e.getMessage)
         }
         sha1.foreach{
           hash =>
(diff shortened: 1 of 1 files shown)


stage1/Stage1Lib.scala
```diff
@@ -43,7 +43,7 @@ class Stage1Lib( val logger: Logger ) extends BaseLib{
 
   def array2hex(padTo: Int, array: Array[Byte]): String = {
     val hex = new java.math.BigInteger(1, array).toString(16)
-    ("0" * (padTo-hex.size)) ++ hex
+    ("0" * (padTo-hex.length)) ++ hex
```
cvogt: is there any different between the two?

stage1/Stage1Lib.scala
```diff
@@ -69,10 +69,14 @@ class Stage1Lib( val logger: Logger ) extends BaseLib{
...
-        } finally {
-          stream.close()
+        try {
+          try
+            Files.copy(stream, incomplete, StandardCopyOption.REPLACE_EXISTING)
+          finally
+            stream.close()
+        } catch {
+          case e: IOException =>
+            System.err.println(red("downloading failed ") ++ e.getMessage)
```
tim-zh: So what should I do then? Add a retry and print stack trace, or print a stack trace with an info message to mark it as expected error, or leave the code as it is?

stage1/Stage1Lib.scala
tim-zh: A quote from IntelliJ IDEA's inspection:

> While such calls are legitimate, they require an additional implicit conversion to SeqLike to be made.
Considering the common use cases, calling length on arrays and strings may provide significant advantages.

stage1/Stage1Lib.scala
cvogt: So my view is: let's not catch the exception and let the jvm show it to the user. But I want this to be a project where things are openly discussed, so I am open to discuss other opinions.

stage1/Stage1Lib.scala
cvogt: sounds good :)

cvogt:
Btw. re-ran the build successfully. Waiting for a response on my comment.

tim-zh:
PR is closed, since the conclusion is not to catch the exception.

cvogt:
@tim-zh the other minor code improvements would still be nice to merge :)

tim-zh:
@cvogt I'll create another PR as part of #25.

PR #273, opened 2016-10-10
Cleanup file structure, upper case Build.scala, #32

tim-zh:


tim-zh:
There is a problem with producing a warning instead of an error when Build.scala is missing:
in that case BuildBuild.managedBuild has to return something, `null` or `Some[BuildInterface]`. Both options conflict with the rest of the code that relies on BuildBuild.managedBuild.

nailgun_launcher/CbtUrlClassLoader.java
```diff
@@ -44,15 +44,15 @@ public Class loadClass(String name, Boolean resolve) throws ClassNotFoundExcepti
...
       }
     }
   }
-  public CbtURLClassLoader(URL[] urls, ClassLoader parent){
+  public CbtUrlClassLoader(URL[] urls, ClassLoader parent){
     super(urls, parent);
     assertExist(urls);
   }
-  public CbtURLClassLoader(URL[] urls){
+  public CbtUrlClassLoader(URL[] urls){
```
cvogt: Let's try to keep commits focused on one thing please. Renaming build.scala into Build.scala is pretty large and throwing in changes like this make them very hard to discover. Can you break this up into separate commits please?

cvogt:
regarding the warning: let's basically accept build.scala and Build.scala but show a warning by writing to System.err when build.scala is found. Since we'll accept it, we should have something to return, I believe?

tim-zh:
The update takes care of #274.

cvogt:
@tim-zh ah, I missed your updates. Just seeing them now. What do you think about using `cbt-build` instead of `build.cbt`? Might be easier to understand? Also see https://gitter.im/cvogt/cbt?at=58000253891a53016309c461

stage2/BuildBuild.scala
```diff
@@ -20,62 +21,70 @@ trait BuildBuild extends BaseBuild{
 
   override def dependencies =
     super.dependencies :+ context.cbtDependency
-  def managedBuildDirectory: java.io.File = lib.realpath( projectDirectory.parent )
+  def managedBuildDirectory: File = lib.realpath( projectDirectory.parent )
   private object managedBuildCache extends Cache[BuildInterface]
   def managedBuild = managedBuildCache{
     val managedBuildFile = projectDirectory++"/Build.scala"
```
cvogt: Instead of doing the below change, could we simple change this one val instead and show the warning if we are using build.scala here? Maybe easier than the nested if below.

tim-zh: Agreed. I'll make it that way.

PR #289, opened 2016-10-27
Web UI

tim-zh:
Future improvements:

* add dependencies from other sources (github)
* generic way to handle plugins
* browse and clone examples

Appreciate if someone wishes to test it in Safari and Edge.


cvogt:
Very interesting. Just tried it out. Love the dependency search! Fun stuff :)!

I think there is definitely room for a graphical user interface for builds. Shell / text editors are more flexible, but GUIs can make it easier to discover / show what options are there.

Ideally, I'd like to not add more stuff to cbt's core, especially no dependencies. I am working towards modularizing stuff actually. How about we turn this into a plugin or a separate project under the scala-cbt github org, where cbt will eventually move as well?

cvogt:
another option: turning it into a stand-alone tool under the upcoming `tools/` directory.

tim-zh:
I thought about better place for such tool and deliberately made it embedded in CBT. My key point here is convenience.

With CBT UI one needs to type a command (12 chars, though I thought about making it `cbt ui`), wait <1 sec and create a project in one click. That's 3 steps. Adding one extra step make situation worse by 33%. And it also makes it worthless in my eyes.

Finally, I can remove scalaj-http and make http calls manually if you wish. I didn't manage to find a decent alternative to Jetty in terms of dependencies and amount of code. But if you have any proposals I'll replace it too.

stage1/cbt.scala
```diff
@@ -23,6 +23,9 @@ object `package`{
     def parent = lib.realpath(file ++ "/..")
     def string = file.toString
   }
+  implicit class StringExtensionMethods(f: File) {
```
cvogt: name is wrong now

tools/ui/resources/template-project/build/build.scala
```diff
@@ -0,0 +1,5 @@
+import cbt._
+
+class Build(val context: Context) extends BaseBuild##with## {
+
+##projectName####dependencies##}
```
cvogt: Have you seen stage2/Scaffold.scala ? If we would use scala's string interpolation as a template system instead of doing these replacements, we don't have to re-invent the wheel.

tools/ui/resources/template-project/src/main/scala/Main.scala
```diff
@@ -0,0 +1,2 @@
+##package##object Main extends App {
+}
```
cvogt: extends App should not be used. It is buggy and considered deprecated, but I think the Scala team was not clear enough about this so far.

tools/ui/resources/template-project/build/build.scala
tim-zh: I don't see any particular advantages in embedding those templates into `ProjectBuilder`. Besides that, long strings would pollute the code, separate files also offer better experience (like highlighting in text editors).

cvogt:
Nice man, cool stuff :). This is a really cool path to explore. Merging as discussed in private chat.

Just for the record, my (soft, non-blocking) concerns with the current implementation:
* I am worried about inventing our own limited templating solution rather than using something existing, i.e. plain string interpolation as in stage2/Scaffold.scala.
* This is currently limited to generating project files for the first time. Maintenance needs to be done in the text, not via the GUI. This may reduce the usefulness while adding a the second way of starting a project which may fragment users understanding of cbt. Also what the GUI generates exactly is not necessarily obvious to users, but they still need know to maintain. The opportunity however is being able to optimize for different use cases / users.

PR #290, opened 2016-10-30
Move SbtLayout, SonatypeRelease, UberJar plugins to stage2/plugins/

tim-zh:
Since they don't have third party dependencies.


cvogt:
@tim-zh Can't merge this right now. I have some branches that conflict with these changes. Let's delay any refactorings until a later point in time. I'd prefer to limits contributions to implementing new features at this point, not refactorings. Let's discuss on gitter before hand in case you'll  heavily touch (or move) any existing files to avoid conflicts.

PR #298, opened 2016-11-06
Integrate eval

cvogt:
This pr merges the twitter `util-eval` tool into CBT. History has been transferred to see origin and author of the changes in this file, but history has been re-written to eliminate file renames.

Reason why CBT will use this and why the code is moved here rather depended upon: 
- CBT will need a runtime code-generation and execution feature to allow dynamic overriding of methods in order to e.g. set the publish version from the command line
- With minor changes the dependency on twitter `util-core` can be removed, which allows CBT to avoid adding it as a dependency
- According to this PR https://github.com/twitter/util/pull/179, Eval is deprecated within twitter


cvogt:
@fommil This PR copies Apache 2.0 licensed code out of a Twitter code base into CBT's Apache 2.0 licensed code based. I'll add a commit with a Readme attributing this code to twitter-util before the merge. Any ideas what steps might be required to be legally fine? Obviously the authors have not signed CBT's CLA.

fommil:
You need to include the Apache 2.0 and any NOTICE files, and that's it. Since it already comes with a licence (and you can convince yourself that these authors were following reasonable procedures) that's basically the same as a CLA.

cvogt:
recreated as #299, now including missing files

PR #299, opened 2016-11-07
Integrate twitter-eval

cvogt:
This pr merges the twitter `util-eval` tool into CBT. History has been transferred to see origin and author of the changes in this file, but history has been re-written to eliminate file renames.

Reason why CBT will use this and why the code is moved here rather depended upon: 
- CBT will need a runtime code-generation and execution feature to allow dynamic overriding of methods in order to e.g. set the publish version from the command line
- With minor changes the dependency on twitter `util-core` can be removed, which allows CBT to avoid adding it as a dependency
- According to this PR https://github.com/twitter/util/pull/179, Eval is deprecated within twitter


cvogt:
As discussed in #298 no CLA is needed from those committers as the code integrated here is already Apache 2.0 licensed.

cvogt:
FYI @olafurpg Twitter Eval made it into CBT. For Scalatex, but more importantly for overriding build methods dynamically via cli arguments, etc.

Issue #62, opened 2016-03-08
dynamic re-configuration


cvogt:
There are multiple use cases for re-configuring a build at runtime. E.g. appending "-SNAPSHOT" to the version number when deploying a snapshot. Or changing the scalaVersion for cross-builds. It is not obvious how to do this, because CBT builds are classes where things are hard wired after instantiation if members are final. More concretely, this isn't possible:

``` scala
class Build ... {
  def publishSnapshot{
    this.version = this.version + "-SNAPSHOT"
    publish
  }
}
```

Candidate solutions are
1. exposing selected fields like the through the constructor and providing via cli mechanism that fills them in (easy to implement, easy to work with, limited to hard coded use cases)
2. mutable members (easy to implement, tricky to work with, probably not a good idea)
3. providing users with a general purpose runtime mixins mechanism (hard to implement, easy to work with). Could look like this
   
   ``` scala
   class BasicBuild(...){
     def publishSnapshot: Unit = {
       val reconfiguredBuild = this.mixin(
         new BasicBuild{
           override def version = super.version+“-SNAPSHOT”
         }
       ): reconfiguredBuild
       b.publish
     }
   }
   ```
4. Hard-coding something like 3, but specific to CBT (slightly less hard to implement, easy to work with, less universal). It would look at what's given and generate a trait that overrides exactly those fields. E.g. if a version is given via the command line, CBT would create build objects like this
   
   ``` scala
   class DynamicOverrides{
     override def version = super.version+“-SNAPSHOT”
   }
   new Build(...) with DynamicOverrides
   ```
   
   Or if the user needs to override things, there could be an api to do so:
   
   ``` scala
   val reconfiguredBuild = this.replace( version = this.version+"-SNAPSHOT" )
   ```
   
   or
   
   ``` scala
   val reconfiguredBuild = (this.version = this.version+"-SNAPSHOT")
   ```
   
   or
   
   ``` scala
   val reconfiguredBuild = this.version.set(this.version+"-SNAPSHOT")
   ```
   
   All of these would need to be magical of course, turning the given expression into the right hand side of a def of a mixed in trait.
5. On the Typelevel front people are considering creating an alternative frontend (which CBT supports), that is not based on inheritance, but a state monad instead. Would be interesting to see how that would look like and how it would differ from SBT's state monad

cvogt:
so for now I added hard coded solutions for version numbers. We can pick this up again later to make it more general.

[...]

Issue #186, opened 2016-06-26
allow passing additional compiler flags from the command line


cvogt:
e.g. to pass in -Ylog-classpath

cvogt:
fixed by #300

`$ cbt with 'override def scalacOptions = super.scalacOptions :+ "-Xprint:typer"' compile`

PR #300, opened 2016-11-08
Dynamic overrides

cvogt:
this branch will eventually fix https://github.com/cvogt/cbt/issues/62


cvogt:
looks like this fixes https://github.com/cvogt/cbt/issues/186 as well

PR #318, opened 2016-11-14
web gui update

tim-zh:
New features:

* project creation from a bundled example
* browsing example files
* special readme.md treatment


tools/gui/src/Main.scala
```diff
@@ -82,4 +126,35 @@ object Main {
...
+      Files.createDirectories(target.toPath.resolve(source.toPath.relativize(dir)))
+      FileVisitResult.CONTINUE
+    }
+
+    override def visitFile(file: Path, attrs: BasicFileAttributes) = {
+      Files.copy(file, target.toPath.resolve(source.toPath.relativize(file)))
+      FileVisitResult.CONTINUE
+    }
+
+  }
```
cvogt: more fp way of doing this:

```scala
  def listFilesRecursive(f: File): Seq[File] = {
    f +: (
      if( f.isDirectory ) f.listFiles.flatMap(listFilesRecursive).toVector else Seq[File]()
    )
  }

listFilesRecursive(source).sorted.foreach{ f=>
  val t = new File(target, f).toPath
  if( Files.isDirectory(t) ) t.createDirectories
  else Files.copy( new File(source, f).toPath, t  )
}
```

PR #326, opened 2016-11-29
Enhance bash script

cvogt:
fixes bugs / improves usability


cvogt:
also adds dependency on sed. Would be nice if we could get rid of that.

PR #330, opened 2016-12-23
Frege support.

mchav:


cvogt:
would be nice to add a test for this, too that it's compiling fine

Issue #338, opened 2017-02-12
date: invalid option -- 'j'


darkfrog26:
I'm trying to run CBT on Fedora 25 and ended up building nailgun from the source and I'm getting the following output when I attempt to run `cbt`:

```
Stopping background process (nailgun) if running
Compiling cbt/nailgun_launcher
date: invalid option -- 'j'
Try 'date --help' for more information.
touch: invalid date format ‘/home/mhicks/projects/third/cbt/nailgun_launcher/target/scala-2.11/classes/../classes.last-success’
Starting background process (nailgun)
connect: Connection refused
```

darkfrog26:
Nevermind about the connection refused (problem with the ng-server script), but you may want to look into the date formatting issue.

cvogt:
@darkfrog26 you also may want to try `cbt direct <task name>` instead of  `cbt <task name>` if you have trouble with the nailgun-based one.

The date thing, yeah. At some point osx syntax snuck in there. Changed the issue name to reflect that.

PR #341, opened 2017-02-13
Fix date -- invalid option -j

trevorsibanda:
Addresses #338


cvogt:
so I was worried that replacing the date command with an alias may have unexpected side-effects. So I looked up what `-j` does and as it turns out it is not even needed. So I just removed it in https://github.com/cvogt/cbt/pull/342

So I am closing this. Thanks though, much appreciated :)

PR #343, opened 2017-02-13
support for flat classloader and enhanced resources example

cvogt:


cvogt:
fixes https://github.com/cvogt/cbt/pull/343

PR #389, opened 2017-03-06
Fix error with zinc library (wrong md5)

TsoBako:


cvogt:
awesome, thx so much :)

Issue #419, opened 2017-03-12
be smarter about looping for errors in build files and cbt itself


cvogt:
Every build should loop if it's managedBuild does not succeed. This may require marrying loadRoot and callReflective.

Right now we only loop correctly if a build once succeeded. In particular looping over a project's sources works correctly, while looping over builds does not recognize new files while a build does not compile.

cvogt:
This means we may not even have to collect the watch files all over the place, but could have many small watchers in multiple places.

It sounds like probably we should propagate task calling simply forwards across builds instead of the `finalBuild` mechanism

cvogt:
a many small watchers is not going to work, because a watcher blocks. so aggregating trigger files as we go along anyways, but cascadingly larger watcher as we succeed deeper into the build.

Issue #406, opened 2017-03-08
Feature request: simple way to clear screen on every loop iteration


olafurpg:
In sbt, I usually set `triggeredMessage in ThisBuild := Watched.clearWhenTriggered`. This triggers `clear` to run on very file-save event before running compile/run/test. `clear` clears the console screen, so it's easy to distinguish console output between each run. It would be nice if it was easy/automatic to do something similar in cbt.

EDIT, I couldn't find any mention of "clear" in the readme or issue, so I opened this ticket. Maybe it's documented somewhere :/

cvogt:
Documentation is scarce right now :). Looks like an ansi code does the job.
This clears the screen for me

cvogt:
Support being added in https://github.com/cvogt/cbt/pull/420 as
```
override def loop{
  lib.clearScreen
  super.loop
}
```

Issue #405, opened 2017-03-08
cbt loop run does not re-compile on file save


olafurpg:
```
cbt tools createMain
cbt loop run
// file save triggers recompile + run
// save again with edit and no recompile, only run on stale file
```
I expected the the second save to trigger a recompile and run on the new file sources.

Note, it's possible to workaround this by Ctrl-c to exit `loop run` and restart `cbt loop run`.

cvogt:
Please try this branch: https://github.com/cvogt/cbt/pull/420

Issue #202, opened 2016-07-06
loop command does not trigger initial execution


benjaminfrank:
`cbt loop run` does not run until a file changes. probably a simple `call(build)` in Stage2 after registering the watch service might be enough to solve this.
Or maybe this is intended behaviour, not sure :)

PR #420, opened 2017-03-12
revamp loop feature

cvogt:
now CBT and builds pass their file names to the current build via the context. The build then simply blocks until any file changes. Then it returns with a special exit code, which the bash script picks up and restarts CBT. Thats works well for looping over project files. It works less well for looping over builds and CBT itself. For this a build has to success once, so that the .cbt-loop.tmp file exists. Then looping works for cbt and builds, but the file list is not updated in case of compile errors, etc.

Fixes
- https://github.com/cvogt/cbt/issues/406
- https://github.com/cvogt/cbt/issues/405
- https://github.com/cvogt/cbt/issues/202
- https://github.com/cvogt/cbt/issues/50
- https://github.com/cvogt/cbt/issues/22

We should improve for 1.0 in https://github.com/cvogt/cbt/issues/419
to handle looping over build files and cbt itself smarter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ node_modules
 examples/dotty-example/_site
 examples/scalajs-plain-example/server/public/generated
 examples/scalajs-react-example/server/public/generated
+.cbt-loop.tmp
--- a/cbt
+++ b/cbt
@@ -199,6 +199,7 @@ stage1 () {
 			log "Running JVM directly" "$@"
 			# JVM options to improve startup time. See https://github.com/cvogt/cbt/pull/262
 			java $JAVA_OPTS $DEBUG -Xmx6072m -Xss10M -XX:MaxJavaStackTraceDepth=-1 -XX:+TieredCompilation -XX:TieredStopAtLevel=1 -Xverify:none -cp $NAILGUN$TARGET cbt.NailgunLauncher $(time_taken) "$CWD" "$@"
+			exitCode=$?
 		else
 			log "Running via background process (nailgun)" "$@"
 			for i in 0 1 2 3 4 5 6 7 8 9; do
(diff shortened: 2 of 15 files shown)


olafurpg:
I tried this out and bumped into a few issues, steps to reproduce

```
mkdir examples/loop
cd examples/loop
cbt tools createMain
cbt loop run
# Everything great so far, edits to Main.scala trigger recompile+run
# In separate terminal tab, while `cbt loop run` is still running in first tab
cbt loop createBuild
# `cbt loop run` stops, and when I try to restart it fails with error:
java.lang.AssertionError: Assertion failed: environment variable CBT_LOOP_FILE not defined
	at cbt.Stage0Lib._assert(Stage0Lib.java:19)
	at cbt.NailgunLauncher.main(NailgunLauncher.java:89)
```

I couldn't run `cbt` in any project until I killed the pid listening to port 4444.

After killing cbt, I ran `cbt loop run` again in `examples/loop` and when I edited `loop/build/build.scala`, I got a NPE

```
java.lang.NullPointerException
	at cbt.BaseBuild$class.loop(BasicBuild.scala:335)
	at loop_build.Build.loop(build.scala:3)
```

The handling of `CBT_LOOP_FILE` seems to be the root cause again.

cvogt:
@olafurpg fixed it, I believe. I had missed something with the background process. It's always a bit tricky with the background process. It captures a lot of stuff when you first start it up, e.g. environment variables. And then those stick around as a side-effect making debugging harder.

olafurpg:
I can confirm that running `cbt tools createBuild` while running `cbt loop run` worked now. Thank you for fixing that! I continued experimenting and bumped into another issue

```
cbt loop run # in project with Main.scala and build/build.scala
# add dependency in build/build.scala, save, Main.scala runs again, all good
# edit Main.scala to use new dependency, save and get following error:
/Users/ollie/dotfiles/bin/cbt: line 168: 93014 Abort trap: 6           $NG_SERVER 127.0.0.1:$NAILGUN_PORT >> $nailgun_out 2>> $nailgun_err
# Exit code: 227
```

I reran `cbt loop run` and it worked fine again, but found another issue with removing dependencies. This issue is unrelated to `loop`

```
# in project with dependency declared in build.scala and Main.scala using dependency
cbt run # All good
# remove dependency, get following error:
java.lang.NoClassDefFoundError: org/scalafmt/Scalafmt$
```
It seems I need to run `cbt direct clean` to be able to run `cbt run` again. I guess the zinc cache should be invalidated when a dependency is removed (or when build.scala changes, because you could have changed scalacOptions).

cvogt:
@olafurpg The nailgun background process is still not 100% reliable. Try `cbt direct <task>` instead to circumvent it, whenever you encounter a problem with `cbt <task>`. Occasionally killing the background process helps `cbt kill` or `killall -KILL java`.

If you find way to reproduce problems, so I can debug them that would be great.

This bug with zinc is unexpected, I'll check it out.

olafurpg:
> If you find way to reproduce problems, so I can debug them that would be great.

I'll report again if I can reproduce that 227 exit code.

cvogt:
@olafurpg fir for the zinc problem in https://github.com/cvogt/cbt/pull/424

PR #429, opened 2017-03-13
fix file watching for real

### LINKED ISSUE #418: re-run task when looping on enter

### LINKED ISSUE #419: be smarter about looping for errors in build files and cbt itself

### LINKED ISSUE #428: nailgun error
@cvogt To reproduce nailgun error:

```
cbt loop run 
# Runs OK
# Remove dependency, compile error as expected
# Add dependency again, get error:
 cbt: line 273:  1212 Abort trap: 6           $NG_SERVER 127.0.0.1:$NAILGUN_PORT >> $nailgun_out 2>> $nailgun_err
```

You may need to do the remove/save/add/save loop two or three times before the error triggers.

by @olafurpg from https://github.com/cvogt/cbt/commit/eb762476a8a609bf437fc193fd48e6f7d1462730#commitcomment-21278104

cvogt:
last file watching update didn’t work well enough. This now
- rips out barbary watch service as it seems buggy crashing the jvm
- make cbt exclusively write files to watch to a file
- uses fswatch instead watching all files in that file

fixes #428 
fixes #419 
fixes #418

### DIFF
--- a/DEVELOPER_GUIDE.txt
+++ b/DEVELOPER_GUIDE.txt
@@ -29,7 +29,7 @@ compatibility/      Java interfaces that all CBT versions are source compatible 
 nailgun_launcher/   Self-contained helper that allows using Nailgun with minimal permanent classpath. (Is this actually needed?)
 realpath/           Self-contained realpath source code to correctly figure our CBTs home directory. (Open for replacement ideas.)
 stage1/             CBT's code that only relies only on Scala/Java built-ins. Contains a Maven resolver to download libs for stage2.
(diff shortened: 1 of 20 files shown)


cvogt:
this also get's rid of cbt's second last dependency barbary watch service, which I know @fommil will like ;). Only jgit left.

PR #443, opened 2017-03-18
Scalafix plugin

darthorimar:


olafurpg:
Cool! Small heads up, the scalafix integration will soon change from scalafix-nsc to scalahost (in scalameta) + scalafix-cli.

cvogt:
@darthorimar Great stuff :)! Mind adding a test? E.g. that before there is no @volatine but after there is? The tests in `test.scala` are not super organized right now, but shouldn't be too hard to follow.

@olafurpg do you have an example somewhere? I saw some snippets in the docs, but something that e.g. shows how to pass options when using scalafix as a library would be nice.

olafurpg:
I'm referring to changes that have not been merged into scalafix. We are moving to sbt-scalahost (see scalameta repo and scalameta/sbt-semantic-example), so the eventual scalafix integration will be closer to that then the current sbt-scalafix. Scalafix is still evolving a lot in terms of integrations, so you might want to wait a bit with this pr until it stabilizes.

test/test.scala
```diff
@@ -128,6 +128,9 @@ object Main{
     implicit val classLoaderCache: ClassLoaderCache = new ClassLoaderCache( new java.util.HashMap )
     def Resolver(urls: URL*) = MavenResolver(cbtLastModified, mavenCache, urls: _*)
 
+    def readFile(name: String) =
+       scala.io.Source.fromFile(cbtHome ++ name).mkString
```
cvogt: there is actually `lib.read(File):String` already

test/test.scala
```diff
@@ -381,6 +386,17 @@ object Main{
       assert(res.out.contains("via parent.parent: true 2"), res.out)
     }
 
+    {
+      runCbt("../examples/scalafix-example", Seq("clean", "force"))
```
cvogt: is this needed? Otherwise let's not do it. `clean` always comes with a risk.

test/test.scala
```diff
@@ -381,6 +386,17 @@ object Main{
...
+      val res = runCbt("../examples/scalafix-example", Seq("compile"))
+      assert(res.exit0)
+      val source = readFile("/examples/scalafix-example/Main.scala")
```
cvogt: let's use `cbtHome / "examples" / "scalafix-example" / "Main.scala"` here.

test/test.scala
```diff
@@ -381,6 +386,17 @@ object Main{
...
+      assert(source contains "@volatile")
+      assert(source contains ": Unit")
+      assert(source contains ": String ")
+      assert(source contains "import scala.collection.immutable")
```
cvogt: can we undo the file changes after the test, so that next time the test just passes because it did once in the past and so they don't show up in git?

cvogt:
@olafurpg This is part of a GSOC qualification exercise, so let's finalize this as is, merge it and upgrade whenever scalafix is ready for it.

test/test.scala
darthorimar: Yes, otherwise it would not call compile method if the scalafix-example was already compiled before. So, the code reformating won't be done too.

cvogt: A better way to do this is making sure the last modified timestamp of the file is newer than when you last compiled. This probably goes along well with resetting the file contents. Otherwise this should help: `Files.setLastModifiedTime( ..., FileTime.fromMillis(System.currentTimeMillis) )`

cvogt: (CBT bases it's compile caching on the last modified timestamps of the source files)

test/test.scala
darthorimar: I can not find this method in Lib.scala. Where should it be?

cvogt: ah, sorry. it is `file.readAsString` (in `cbt.scala`)

test/test.scala
```diff
@@ -381,6 +387,19 @@ object Main{
...
+      val sourceFile = cbtHome / "examples" / "scalafix-example" / "Main.scala"
+      val sourceBefore = sourceFile.readAsString
+      val res = runCbt("../examples/scalafix-example", Seq("compile"))
+      assert(res.exit0)      
+      val sourceAfter = sourceFile.readAsString
+      assert(sourceAfter contains "@volatile")
+      assert(sourceAfter contains ": Unit")
+      assert(sourceAfter contains ": String ")
+      assert(sourceAfter contains "import scala.collection.immutable")
+      lib.write(sourceFile, sourceBefore, StandardOpenOption.TRUNCATE_EXISTING)
```
cvogt: as per java documentation
> if no options are present then this method works as if the CREATE, TRUNCATE_EXISTING, and WRITE options are present

cvogt: so no need to pass this.

plugins/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,28 @@
...
+object Scalafix {
+  def scalacOptions( rootPath: Path, nscJar: File ) =
+    Seq(
+      "-Xplugin:" ++ nscJar.string,
+      "-Yrangepos"
+    ) ++ configOption(rootPath)
+
+  def configOption( rootPath: Path ) =
+    Some( rootPath.resolve(".scalafix.conf").toAbsolutePath )
+      .filter(isRegularFile(_)).map("-P:scalafix:" ++ _.toString).toSeq
```
cvogt: This method only adds an option if a file can be found, but otherwise succeeds, but adds nothing. It's better to actually crash in this case and say what file couldn't be found. Otherwise users forget the file or name it the wrong way and they can't figure out why it's not working. Just do `.getOrElse( throw new Exception(...) )` with a useful message.

Also `isRegularFile` prohibits symlinks, how about `.filterNot(isDirectory(_))` instead?

darthorimar: Scalafix has a standard behaviour then no config found

cvogt: I see. ok... fair enough, let's keep it then :).

darthorimar: This won't work because `isRegularFile` also checks that is a real file. Eg., if we pass a wrong file name to the `isDirectory` it just returns `false`

darthorimar: A better check both of them

PR #444, opened 2017-03-18
Scalastyle plugin

cvogt:


plugins/scalastyle/Scalatyle.scala
```diff
@@ -0,0 +1,98 @@
...
+trait Scalastyle extends BaseBuild {
+  def scalastyle = {
+      val dirList = List(projectDirectory) map (_.getAbsolutePath)
+
+      val pluginHome : String = context.cbtHome.getAbsolutePath + "/plugins/" + "scalastyle"
+
+	    val result = ScalaStyle(dirList, ScalaStyle.checkConfig(projectDirectory.getAbsolutePath, Some(pluginHome))) 
+      result match {
+        case 1 => println("Error Scalastyle Checking")
+        case 2 => println("Config file for ScalaStyle not found")
```
cvogt: also please don't use primitive types (or string) to encode a multiple-choice situation. If you really need this use an Option, a boolean or a sealed trait with case objects. But we might get away without it entirely here. Why don't we move this into the object, instead of having it in this trait? The trait should be only wiring-up, no logic.

plugins/scalastyle/Scalatyle.scala
```diff
@@ -0,0 +1,98 @@
...
+    
+    def checkConfigExists(filename: String) : Option[String] = {
+      val file = new File(filename)
+      if (file.exists()) Some(filename) else None
+    }
+
+    checkConfigExists(directory + defaultConfigName) orElse (
+      fallback flatMap (fdir => checkConfigExists(fdir + defaultConfigName))
+    )
+  }
```
cvogt: configs probably make most sense on a per-project bases, not in the user's home directory. Can we default to the project's directory and give the user the option to choose his own config file location?

test/test.scala
```diff
@@ -223,6 +223,7 @@ object Main{
     compile("../examples/scalafmt-example")
     compile("../examples/scalariform-example")
     compile("../examples/scalatest-example")
+    compile("../plugins/scalastyle")
```
cvogt: Please also add an example project and a test that checks that Scalastyle is warning about the right things there.

plugins/scalastyle/Scalatyle.scala
```diff
@@ -0,0 +1,98 @@
...
+
+
+
+  def apply(directories: List[String], configFile: Option[String]) : Int = {
+    val exitVal = configFile match {
+      case Some(_) => {
+         val conf = MainConfig(false).copy(config = configFile, directories = directories)
+         if (conf.error) {
+            1
```
cvogt: you just set it to false in line 53, so I doubt it could be true, here.

plugins/scalastyle/Scalatyle.scala
```diff
@@ -0,0 +1,98 @@
...
+      fallback flatMap (fdir => checkConfigExists(fdir + defaultConfigName))
+    )
+  }
+
+
+
+  def apply(directories: List[String], configFile: Option[String]) : Int = {
+    val exitVal = configFile match {
+      case Some(_) => {
+         val conf = MainConfig(false).copy(config = configFile, directories = directories)
```
cvogt: why .copy and not pass them as arguments?

plugins/scalastyle/Scalatyle.scala
```diff
@@ -0,0 +1,98 @@
...
+         if (conf.error) {
+            1
+         } else {
+            if (execute(conf)) 1 else 0
+         }
+      }
+      case None => {
+        2
```
cvogt: this tells the caller of the function something he already knows. If he passes None he'll always gets 2. Not that useful to the caller. How about we make `configFile` non optional instead?

cvogt: or even better, let the user create the MainConfig object. This apply method seems to be doing nothing interesting instead of calling `execute`. If we let users call that code directly, the code becomes simpler.

plugins/scalastyle/Scalatyle.scala
```diff
@@ -0,0 +1,98 @@
...
+      } 
+    }
+   
+    exitVal
+  }
+
+
+  private[this] def now(): Long = new Date().getTime()
```
cvogt: Let's make this a local function inside of execute. It's not needed outside. Principle of least power :). Expose the fewest things necessary.

plugins/scalastyle/Scalatyle.scala
```diff
@@ -0,0 +1,98 @@
...
+
+  private[this] def execute(mc: MainConfig)(implicit codec: Codec): Boolean = {
```
cvogt: now that I read this function I realize, we are never actually passing MainConfig anywhere, are we? We just use it as a container we read outselves. In that case let's not use it at all. Let's instead follow the nested case class pattern described in the plugin author guide https://github.com/cvogt/cbt/blob/master/doc/plugin-author-guide.md and e.g. implemented by https://github.com/cvogt/cbt/blob/master/plugins/scalafmt/Scalafmt.scala or https://github.com/cvogt/cbt/blob/master/libraries/proguard/Proguard.scala

plugins/scalastyle/Scalatyle.scala
```diff
@@ -0,0 +1,98 @@
...
+    val start = now()
+    val configuration = ScalastyleConfiguration.readFromXml(mc.config.get)
```
cvogt: Ah, I see. So Scalastyle reads its config xml files into a `ScalastyleConfiguration` class. You know what? Let's remove anything related to xml from your plugin entirely and let users just provide a `ScalastyleConfiguration` instance. If they want to use an xml file they can just call the `readFromXml` method themselves. The plugin doesn't need to be tied to xml at all.

plugins/scalastyle/Scalatyle.scala
```diff
@@ -0,0 +1,98 @@
...
+    val cl = mc.externalJar.flatMap(j => Some(new java.net.URLClassLoader(Array(new java.io.File(j).toURI().toURL()))))
```
cvogt: there is no need for creating a classloader yourself. I presume you need the classloader of the project that's being checked? there is a classLoader method in BaseBuild that gives you access to it, you just need to pass it all the way in here.

plugins/scalastyle/Scalatyle.scala
```diff
@@ -0,0 +1,98 @@
...
+        val encoding = mc.xmlEncoding.getOrElse(codec.charSet).toString
+        XmlOutput.save(config, x, encoding, messages)
+      }
+      case None =>
+    }
+
+    if (!mc.quiet) println("Processed " + outputResult.files + " file(s)")
+    if (!mc.quiet) println("Found " + outputResult.errors + " errors")
+    if (!mc.quiet) println("Found " + outputResult.warnings + " warnings")
+    if (!mc.quiet) println("Finished in " + (now - start) + " ms")
```
cvogt: Other cbt plugins use a logger for this. Let's keep things the same.

plugins/scalastyle/Scalatyle.scala
```diff
@@ -0,0 +1,98 @@
...
+    // scalastyle:off regex
+    val config = ConfigFactory.load(cl.getOrElse(this.getClass().getClassLoader()))
+    val outputResult = new TextOutput(config, mc.verbose, mc.quiet).output(messages)
+    mc.xmlFile match {
+      case Some(x) => {
+        val encoding = mc.xmlEncoding.getOrElse(codec.charSet).toString
+        XmlOutput.save(config, x, encoding, messages)
+      }
+      case None =>
+    }
```
cvogt: how about we show the errors to System.err instead of writing them to a file? More convenient.

plugins/scalastyle/scalastyle_config.xml
```diff
@@ -0,0 +1,142 @@
+<scalastyle commentFilter="enabled">
```
cvogt: Do we need this whole file here? Can we just write a ScalastyleConfiguration with a small number of rules in an example instead?

cvogt:
I cleaned this up yesterday, trimmed it down slightly and created a PR. Closed in favor of https://github.com/cvogt/cbt/pull/479

PR #459, opened 2017-03-25
Find nailgun-server jar on Debian

jodersky:
Installing nailgun via apt on Debian copies the server jar to
/usr/share/java/nailgun-server.jar. This change enables the sbt
launcher to find the jar.


cvogt:
Great :)!

PR #462, opened 2017-03-27
scala meta and scalafix

cvogt:


cvogt:
cc @olafurpg @shanedelmore

plugins/scalafix/Scalafix.scala
```diff
@@ -1,28 +1,64 @@
...
+          files.map(_._1).mkString(File.pathSeparator)
+        )
+
+      def context(file: File): ( RewriteCtx[Mirror], RewriteCtx[ScalafixMirror] ) = (
+        scalafix.rewrite.RewriteCtx(
+          mirror.dialect(file).parse[Source].get, ScalafixConfig(), mirror
+        ),
+        scalafix.rewrite.RewriteCtx(
+          mirror.dialect(file).parse[Source].get, ScalafixConfig(), ScalafixMirror.fromMirror( mirror )
+        )
```
cvogt: @olafurpg tried my best, but this currently fails with
java.lang.UnsupportedOperationException: scalafix.rewrite.ScalafixMirror.fromMirror $anon#typeSignature requires the semantic api

Is that what you mentioned, that you can't trigger the built-in plugins via scalahost?

olafurpg: ExplicitImplicit does not yet work with scala.meta.Mirror, the `ScalafixMirror.fromMirror` method creates a dummy implementation of `ScalafixMirror.typeSignature` which throws this error.

You can disable `ExplicitImplicit` to get it to work.

The current design is a bit in flux because scalafix-nsc was developed before scalahost/scala.meta semantic api existed. My plan is to replace scalafix-nsc/ScalafixMirror entirely with scalahost's "offline" mirror. That is, the "offline" mirror is persisted during compilation by the scalahost compiler plugin and it can be loaded outside of the compilation pipeline by tools like scalafix or code search engines. However, the scalahost mirror currently only support one method: `Ref.symbol` (name resolution). `Symbol.members` and `Term.tpe` is on the roadmap.

ShaneDelmore: @olafurpg Does this mean none of the semantic rewrites will work, or is it something particular to ExplicitImplicit?  I wanted to update the example with a few type replacements or additions, for example if I find Either.map in 2.11.8 then add a lib that provides an enrichment for map on Either.

olafurpg: Particular to ExplicitImplicit, because it uses the `typeSignatures` method in `ScalafixMirror`. Everything in the `Mirror` api, like `Ref.symbol`, will work.

ShaneDelmore: That's great news, thanks.

PR #466, opened 2017-03-28
WIP - adding example to cross-rewrite

ShaneDelmore:
This is not even close to ready, just making a pr to see if I am on the right track.


examples/cross-rewrite-example/build/build.scala
```diff
@@ -6,67 +6,79 @@ import scalafix.util._
 import scalafix.util.TreePatch._
 import scalafix.util.TokenPatch._
 
-class Build(val context: Context) extends BaseBuild{ outer =>
-  override def defaultScalaVersion = "2.11.8"
+class Build(val context: Context) extends BaseBuild { outer =>
+  override def defaultScalaVersion = "2.12.1"
```
ShaneDelmore: I changed the default because I thought people might like to write code in the latest version and backwards compile.

examples/cross-rewrite-example/build/build.scala
```diff
@@ -6,67 +6,79 @@ import scalafix.util._
...
+
+  override def test: Dependency = {
+    new BasicBuild(context) with ScalaTest {
+      override def dependencies = outer +: super.dependencies
+      override def defaultScalaVersion = "2.12.1"
```
ShaneDelmore: Is this correct that I need to override version for test build (why would I ever want to test on a different version?)
Or should my test be extending my main project somehow to have access to the code in it?

examples/cross-rewrite-example/build/build.scala
```diff
@@ -6,67 +6,79 @@ import scalafix.util._
...
+            override def defaultScalaVersion = v
+            override def dependencies =
+              super.dependencies ++ Resolver(mavenCentral).bind(dep)
+            override def projectDirectory = d
+            override def scaladoc = None
+            override def sources = {
+              val fromTo = lib.autoRelative(outer.sources).collect {
+                case (location, relative) if location.isFile =>
+                  location -> projectDirectory / "src" / relative
+              }
```
ShaneDelmore: scalafmt must have changed all of this slightly.  Would you be interested in a pr adding a .scalafmt config to the project?

examples/cross-rewrite-example/src/Main.scala
```diff
@@ -1 +1,6 @@
-import scala.concurrent.Future
+object Main {
+  def main(args: Array[String]): Unit = {
+    //For 2.11.8 rewrite to case match
+    Disjunction.intOrString.map(println)
```
ShaneDelmore: I could either rewrite this to a case statement for scalaz 2.11.8 (I don't believe they have an either enrichment for map) or I could change the type to \/.  Either one would require the semantic mirror which I am not clear is actually available based on @olafurpg comment about ExplicitImplicit being available.  I will experiment more once I get tests to pass.

examples/cross-rewrite-example/test/disjunction_spec.scala
```diff
@@ -0,0 +1,10 @@
+import collection.mutable.Stack
+import org.scalatest._
+
+class Test extends FunSpec with Matchers {
```
ShaneDelmore: This definitely does not work, I need to dig more to understand why.

plugins/scalatest/ScalaTest.scala
```diff
@@ -9,7 +9,7 @@ trait ScalaTest extends BaseBuild{
     runSuites( suiteNames.map( loadSuite( _, classLoader ) ) )
     ExitCode.Success
   }
-  override def dependencies = super.dependencies ++ Resolver( mavenCentral ).bind( ScalaDependency("org.scalatest","scalatest","2.2.4") )
+  override def dependencies = super.dependencies ++ Resolver( mavenCentral ).bind( ScalaDependency("org.scalatest","scalatest","3.0.1") )
```
ShaneDelmore: There is no 2.x version of Scalatest for scala 2.12 so I bumped to scalatest 3.0.1.

plugins/scalatest/build/build.scala
```diff
@@ -5,6 +5,6 @@ class Build(val context: Context) extends BaseBuild{
     super.dependencies
     :+ context.cbtDependency
   ) ++ Resolver( mavenCentral ).bind(
-    ScalaDependency("org.scalatest","scalatest","2.2.4")
+    ScalaDependency("org.scalatest","scalatest","3.0.1")
```
ShaneDelmore: I haven't looked into this yet but don't like that there are these duplicated strings in multiple files.  Maybe I need to create a Dependencies.scala file to share them.

examples/cross-rewrite-example/build/build.scala
cvogt: 👍

examples/cross-rewrite-example/build/build.scala
cvogt: you overwrote defautScalaVersion is class `Build`, but here you are instantiating class `BasicBuild`. That's fine, but it doesn't magically pick up on what you overwrote in `Build`. So if you want this as 2.12.1 you need to provide that info here somehow.

examples/cross-rewrite-example/build/build.scala
cvogt: Not completely warmed up to scalafmt yet. I played around with it a bit, but it currently does not allow me to produce code that I find as readable as scalariform. So for now I have settled on Scalariform for CBT and there is a preconfigured `plugins.scalariform` plugin you can add as a dependency to your BuildBuild, allowing you to mix in `Scalariform` into your build, which adds a preconfigured task `scalariform`, which by default is not run on compile, but can be easily added (in which case one has to call scalariform.apply there, because it is lazy).

ShaneDelmore: No problem.  I don't mind using scalariform but I may take a shot at making a .scalafmt.conf that matches your scalariform preferences just to see if I can.

cvogt:
the tests check for the existence of a cats AND scalaz crossbuilds, which is why this still fails ;).

ShaneDelmore:
Ah, that makes sense. That should be easy enough for me to resolve Saturday.

examples/cross-rewrite-example/build/build.scala
```diff
@@ -28,22 +28,25 @@ class Build(val context: Context) extends BaseBuild { outer =>
...
+      MavenDependency("org.scalaz", "scalaz-core", "7.2.10"),
+      Seq(
+        //This works
+        AddGlobalImport(importer"scalaz._"),
+        //None of these work, why?
+        Replace(Symbol("_root_.scala.util.Either."), q"\/"),
+        Replace(Symbol("_root_.Disjunction."), q"Test"),
+        Rename(q"intOrString", q"xor"),
+        //And this works again.  What is different about import rewrites vs the others?
+        RemoveGlobalImport(importer"scala.concurrent.Future")
```
ShaneDelmore: I am not sure if I am missing something or if there is a bug in scalafix, but import releated rewrites work and others do not.  I can't tell what the difference is yet.  I looked back through scalafix code and these are all instances of TreePatch.  Once the patching works this should be as simple as adding scalaz.\/, replacing Either with \/ and Right with \/-

examples/cross-rewrite-example/build/build.scala
```diff
@@ -6,67 +6,95 @@ import scalafix.util._
...
       "scalaz",
-      ScalaDependency( "org.scalaz", "scalaz-core", "7.2.2" ),
+      MavenDependency("org.scalaz", "scalaz-core", "7.2.10"),
       Seq(
+        //This works
+        AddGlobalImport(importer"scalaz._"),
+        //None of these work, why?
+        Replace(Symbol("_root_.scala.util.Either."), q"\/"),
+        Replace(Symbol("_root_.Disjunction."), q"Test"),
+        Rename(q"intOrString", q"xor"),
```
olafurpg: `Rename` should probably be called `RenameThisInstance` (or similar) because it's reference based, I opened https://github.com/scalacenter/scalafix/issues/109 to track that change. To replace call-sites of a specific definition you want to use `Replace`, which is symbol based.

examples/cross-rewrite-example/build/build.scala
```diff
@@ -6,67 +6,95 @@ import scalafix.util._
...
   def libs = Seq[(String, MavenDependency, Seq[Patch])](
     (
       "scalaz",
-      ScalaDependency( "org.scalaz", "scalaz-core", "7.2.2" ),
+      MavenDependency("org.scalaz", "scalaz-core", "7.2.10"),
       Seq(
+        //This works
+        AddGlobalImport(importer"scalaz._"),
+        //None of these work, why?
+        Replace(Symbol("_root_.scala.util.Either."), q"\/"),
```
olafurpg: Have you tried `Mirror.database.toString` to double check that this symbol matches the symbol you want to rewrite? Otherwise, please open an issue in scalafix and I can take a look at it next week.

ShaneDelmore: The type of scala.util.Either is actually scala.package.Either.  Printing out the mirror.database helped, thanks @olafurpg

plugins/scalafix/Scalafix.scala
```diff
@@ -46,19 +50,28 @@ object Scalafix{
...
+          case (from, to) =>
+            implicit val (ctx, ctx2) = context(from)
+            println("printing sources")
+            println(ctx.mirror.sources)
+            println("printing database")
+            println(ctx.mirror.database)
+            println("printing mirror database messages")
+            println(ctx.mirror.database)
+            println("end printing")
+            assert(ctx.mirror.database.toString.trim != "", "I do not believe the database should be empty here")
```
ShaneDelmore: @olafurpg The semantic db appears to be empty here even though sources is not.  Do you think the semantic db is being constructed incorrectly, or that I should open a scalameta issue on this?

plugins/scalafix/Scalafix.scala
```diff
@@ -25,12 +25,16 @@ object Scalafix{
...
+        println("creating semantic mirror")
+        println(s"classpath: ${classpath.string}")
+        println(s"files: ${files.map(_._1).mkString(File.pathSeparator)}")
         Mirror(
           classpath.string,
           files.map(_._1).mkString(File.pathSeparator)
         )
+      }
 
       def context(file: File): ( RewriteCtx[Mirror], RewriteCtx[ScalafixMirror] ) = (
```
olafurpg: `RewriteCtx[ScalafixMirror]` is not necessary here, `ScalafixMirror` will go away in favor of `Mirror`.

examples/cross-rewrite-example/build/build.scala
```diff
@@ -6,67 +6,101 @@ import scalafix.util._
 import scalafix.util.TreePatch._
 import scalafix.util.TokenPatch._
 
-class Build(val context: Context) extends BaseBuild{ outer =>
+class Build(val context: Context) extends BaseBuild { outer =>
```
olafurpg: There's missing a `with Scalameta` here to inject the scalahost compiler plugin, which generates `semanticdb` files under `target/`.

examples/cross-rewrite-example/build/build.scala
olafurpg: Scalafmt does not have any settings which mimic scalariform, they're very different formatters.

plugins/scalafix/Scalafix.scala
olafurpg: The project source files need to be compiled with the scalahost compiler plugin for the database to be built, so the `Scalameta` plugin needs to be mixed in the root build. The mirror is composed of a `classpath` (pointing to at least one directory containing `semanticdb` files) and a `sourcepath` (pointing to scala source files). In this case, the classpath does not contain any semanticdb files, maybe the Mirror constructor should fail in that case.

examples/cross-rewrite-example/build/build.scala
ShaneDelmore: Doh!  Thanks for catching that.

plugins/scalafix/Scalafix.scala
ShaneDelmore: I do believe fail fast would be better here, but this should be enough to advice to get me back on track. 👍

examples/cross-rewrite-example/build/build.scala
```diff
@@ -6,11 +6,48 @@ import scalafix.util._
...
+      }
+
+    def setVersion(version: String): Build => Build = _.copy(scalaVersion = version)
+    def addDependency(dep: MavenDependency): Build => Build = _.copy(dependencies = dependencies :+ dep)
+    def removeDependency(dep: MavenDependency): Build => Build = _.copy(dependencies = dependencies :- dep)
+
+    //Over time I would hope to build up a community libraryof rewrites so many of the transforms would be pulled from
+    //a library instead of declared in the build
+  }
+
```
ShaneDelmore: Ignore everything below here, I didn't mean to commit it.

ShaneDelmore:
I forgot all about this PR.  Will try to take a look at it in a few days and see if I can figure out why tests fail.

PR #472, opened 2017-03-30
Scalatest for 2.12

cvogt:


libraries/scalatest-runner/Runner.scala
```diff
@@ -0,0 +1,37 @@
...
+      _.execute(testName, configMap, color, durations, shortstacks, fullstacks, stats)
+    }
+  }
+
+  def discoverSuites( discoveryPath: File, classLoader: ClassLoader ): Seq[String] = {
+    classLoader
+      .loadClass("org.scalatest.tools.SuiteDiscoveryHelper")
+      .getMethod("discoverSuiteNames", classOf[List[_]], classOf[ClassLoader], classOf[Option[_]])
+      .invoke(null, List(discoveryPath.toString ++ "/"), classLoader, None)
+      .asInstanceOf[Set[String]]
```
ShaneDelmore: This really takes the magic out of test discovery. I had wondered how that worked.

libraries/scalatest-runner/build/build.scala
```diff
@@ -0,0 +1,8 @@
+import cbt._
+
+class Build(val context: Context) extends BaseBuild{
+  override def dependencies = super.dependencies ++
+    Resolver( mavenCentral ).bind(
+      ScalaDependency("org.scalatest","scalatest", if(scalaMajorVersion == "2.12") "3.0.1" else "2.2.6")
```
ShaneDelmore: Why not 3.0.1 for all versions?

libraries/scalatest-runner/Runner.scala
cvogt: Took me forever to find this back when I did :).

libraries/scalatest-runner/build/build.scala
cvogt: Hadn't thought about it too much, but mainly because I am running tests coded against 2.2.x with this. If I am not mistaken providing scalatest 3.x on the classpath in 2.11 should allow you to run 3.x based tests with this as well. Now that this is running as part of the test project's code instead of the build, we may actually be able to get rid of the reflection for discovery here. But later.

PR #487, opened 2017-04-11
Fix ScalaPB options definition

atvictor:
ScalaPB options were never used and were not in underscore case.
It has been tested with no options and it works (i.e. `--scala_out=<empty>:<path>`).


cvogt:
great stuff :)!

PR #514, opened 2017-06-12
Make scalameta-version configurable

megri:
May need some tests but for now it downloads the specified version of scalameta as long as it's available ;)


cvogt:
LGTM

PR #515, opened 2017-06-12
Rudimentary Macroparadise plugin

megri:


cvogt:
love it. Ideal would be a follow up commit or PR, that adds your experiment to the `examples/` folder

megri:
Is it ok in its raw adulterated form as shown on megri/cbt-test?

cvogt:
yes. ideally add a line here to check it produces the right output: https://github.com/cvogt/cbt/blob/master/test/test.scala#L424

and if you want to turn it into a single build file multi-project build, that might be nice :)

cvogt:
Gave it another look. How is this different from the existing Scala paradise plugin?

cvogt:
Except for better implementation and version number?

cvogt:
Should it maybe replace it?

megri:
I believe Scalaparadise.scala enables scala.reflection-based (old-style) macros for Scala 2.10-projects and macro annotations for 2.11+.

Macroparadise.scala on the other hand enables scala.meta-style macros. I believe they are currently limited to macro annotations only.

megri:
Updated the PR to use a single build.scala. Please review and comment on whether or not it looks sane :)

megri:
Rebased PR on upstream

megri:
Added the renaming as a separate commit. Can squash if desired

cvogt:
Awesome. Merging

PR #516, opened 2017-06-13
Fix shellcheck linting warning

megri:


cvogt:
ah, was it a bug after all?

cvogt:
```
Error: Could not find or load main class  -Xmx1536m -Xss10M -XX:MaxJavaStackTraceDepth=-1 -XX:+TieredCompilation -XX:TieredStopAtLevel=1 -Xverify:none
```
I suppose we have to quote the args containing `=` individually

megri:
Hmm, it "works on my machine"

cvogt:
can you try `cbt direct compile` on your machine?

megri:
So I figured out what's going on.

The original line creates an array: `JAVA_OPTS_CBT=($DEBUG …)`
This is then expanded when calling java with `java … "${JAVA_OPTS_CBT[@]}"`

My change originally replaced the array with an interpolated string — in lack of better bash jargon. Expanding this with the `FOO[@]` syntax seems to work but as the result of that is quoted, what gets passed as an argument is actually something like `java '-Xmx1536m …'` — single-quoted, which in turn passes the whole configuration as a single parameter to java.

Removing the quotes around the variable in the statement seems to have fixed the issue: I can now run both `cbt test` and `cbt direct compile` without issue. My updated PR reflects these changes.

megri:
Argh, no. shellcheck doesn't like this. Gonna give this some more thought.

cvogt:
awesome, thx!

megri:
I ended up wrapping JVM-arguments containing `=` in quotes. This works on my machine (;)) but circle seems to have failed while I was rebasing stuff so I think that's what's being reported.

cvogt:
looks like it passed on circle. nice stuff.

PR #520, opened 2017-06-14
Add tek/splain plugin

megri:


stage2/plugins/SplainPlugin.scala
```diff
@@ -0,0 +1,46 @@
...
+  def scalacOptions( jarPath: File, params: Seq[ParameterPair] ) =
+    Seq( "-Xplugin:" ++ jarPath.string ) ++ mkParamList( params )
+
+  private def mkParamList(params: Seq[ParameterPair]) =
+    params.foldLeft( Seq.empty[String] ) {
+      case (paramList, (param, Some(value))) =>
+        paramList :+ s"-P:splain:$param:$value"
+      case (paramList, (param, None)) =>
+        paramList :+ s"-P:splain:$param"
+    }
```
cvogt: .map seems like the more appropriate operator here. and value.map(":"+_).getOrElse("") would case the second case.

stage2/plugins/SplainPlugin.scala
```diff
@@ -0,0 +1,46 @@
...
+/**
+  * A scala compiler plugin for more concise errors
+  * See: https://github.com/tek/splain
+  */
+trait SplainPlugin extends BaseBuild{
+  import SplainPlugin.ParameterPair
+
+  def splainVersion = "0.2.4"
+
+  def splainArguments: Seq[ParameterPair] = Seq( "all" -> None )
```
cvogt: How about we make a sealed trait with case objects for the cases?

megri: I actually started down that path but wanted some feedback first :)

stage2/plugins/SplainPlugin.scala
megri: Fair enough!

megri: Are you strongly opposed to having two cases? I think it reads a bit cleaner personally

megri:
Please review my changes. 

I tried scoping the members inherited from the trait but they still show up in the list shown when running `cbt` from a project whose build extends the plugin.

megri:
Reading up on access modifiers in Scala, `protected[this]` should make the member accessible only in the inner scope of the extending class. I've created an issue that reflects (heh!) the culprit here: https://github.com/cvogt/cbt/issues/522

PR #511, opened 2017-06-11
Export build information

darthorimar:
Just for review :)

PR #539, opened 2017-06-27
Add support for generating ensime configuration file

### LINKED ISSUE #89: ENSIME plugin?
we'd be happy to support you to write a `.ensime` generator so that people could use cbt with ENSIME.

Some documentation to get you started here http://ensime.github.io/build_tools/

jodersky:
This plugin adds the capability to generate ENSIME configuration files. It's settings are mostly copied from sbt-ensime.

Currently, only configuration for single-project builds is generated. I'll need to look deeper into how I can list a build's DirectoryDependencies, in order add support for multi-module projects.

Closes #89 

FYI @fommil

### DIFF
--- a/stage2/plugins/Ensime.scala
+++ b/stage2/plugins/Ensime.scala
@@ -61,20 +61,7 @@ trait Ensime extends BaseBuild {
     scalaVersion = ensimeScalaVersion,
     javaSources = jdkSource.toSeq,
     javaFlags = ensimeJavaFlags,
-    projects = EnsimeProject(
-      id = EnsimeProjectId(
-        project = name,
-        config = "compile"
-      ),
-      depends = Nil, // currently only one project is supported
-      sources = sources.filter(_.isDirectory),
-      targets = Seq(compileTarget),
-      scalacOptions = scalacOptions.toList,
-      javacOptions = Nil,
-      libraryJars = transitiveDependencies.flatMap(_.exportedClasspathArray).toList,
-      librarySources = resolveWithClassifier(transitiveDependencies, Classifier.sources),
-      libraryDocs = resolveWithClassifier(transitiveDependencies, Classifier.javadoc)
-    ) :: Nil
+    projects = findProjects(this)
   )
 
   /** Generate an ENSIME configuration file for this build. */
@@ -169,6 +156,30 @@ object Ensime {
             s"ensime: could not find $classifierName of ${dependency.mavenDependency.serialize}")
           Seq.empty
       }
+    }
+  }
+
+  /** Find and convert all (transitive) dependencies of a build to a sequence of ENSIME projects. */
+  def findProjects(root: BaseBuild)(implicit logger: Logger, cache: JMap[AnyRef, AnyRef], classLoaderCache: ClassLoaderCache): Seq[EnsimeProject] = {
+    def asProject(base: BaseBuild) = EnsimeProject(
+      id = EnsimeProjectId(
+        project = base.name,
+        config = "compile"
+      ),
+      depends = base.transitiveDependencies.collect{
+        case b: BaseBuild => EnsimeProjectId(b.name, "compile")
+      },
+      sources = base.sources.filter(_.isDirectory),
+      targets = Seq(base.compileTarget),
+      scalacOptions = base.scalacOptions.toList,
+      javacOptions = Nil, // TODO get javac options from base build
+      libraryJars = base.transitiveDependencies.flatMap(_.exportedClasspathArray).toList,
+      librarySources = resolveWithClassifier(base.transitiveDependencies, Classifier.sources),
+      libraryDocs = resolveWithClassifier(base.transitiveDependencies, Classifier.javadoc)
+    )
+
+    Seq(asProject(root)) ++ root.transitiveDependencies.collect{ case b: BaseBuild =>
+      asProject(b)
     }
   }
 
@@ -279,7 +290,7 @@ object Ensime {
         "projects" -> conf.projects,
         // subprojects are required for backwards-compatibility with older clients
         // (ensime-server does not require them)
-        "subprojects" -> List(EnsimeModule.fromProjects(conf.projects))
+        "subprojects" -> conf.projects.groupBy(_.id.project).mapValues(EnsimeModule.fromProjects)//conf.projects.flatMap(proj => EnsimeModule.fromProjects(conf.projects))
       ))
       case module: EnsimeModule => sexp(List(
         "name" -> module.name,
--- a/stage2/plugins/Ensime.scala
+++ b/stage2/plugins/Ensime.scala
@@ -21,9 +21,6 @@ import scala.util.control.NonFatal
   */
 trait Ensime extends BaseBuild {
 
-  /** The version of ensime to use. */
-  def ensimeServerVersion: String = "2.0.0-SNAPSHOT"
-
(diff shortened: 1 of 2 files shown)


jodersky:
oooh, I just saw #511. It might help to get multi-module builds working

fommil:
Nice! Don't forget to include the NOTICE file from ensime-sbt to stay compliant with the license.

Is there any way to add tests in cbt yet, like sbt scripted?

jodersky:
@fommil, I'm not sure how a NOTICE file is usually structured. Is the one I added acceptable?

Regarding tests, we can probably add a test project to /tests or in a sub directory of /stage2/plugings. @cvogt what's the preferred way of adding tests to plugins?

cvogt:
Woo, this is really awesome :D.

Regarding tests. Right now everything is triggers from `test/test.scala`. For plugins there are usually examples in `examples/ensime/` and those can serve as test cases. You can add something like this towards the end of the file:

```
    {
      val res = runCbt("../examples/ensime", Seq("ensime"))
      assert(res.exit0)
      assert(res.out contains "<some sexp stuff maybe?>", res.out)
    }
```

checking for some substring might give a good level of confidence already? Wouldn't validate the details obviously.

Regarding the implementation of the plugin. Very clean, I like it. Early this year I start a slightly different plugin code using slightly more unconventional Scala, but solving the problem of the ever-growing number of top-level name-prefixed methods in the Build class body. Instead of having `ensimeScalaJars` a top-level, public method, you could inline it or make it private. Things can still be configured.

Instead of

```
override def ensimeScalaJars = super. ensimeScalaJars :+ ...
```

one would write

```
override def ensimeConfig = super. ensimeConfig.copy(
  scalaJars = super. ensimeConfig.scalaJars :+ ...
)
```

Also instead of having both `ensimeConfig` and `ensime`, I started having only one key, i.e. `ensime`, which is the config case class instance, but giving it an apply method, that then actually does whatever is supposed to happen, i.e. generate the ensime config file. Here is an example: https://github.com/cvogt/cbt/commit/435267714fd9be710945d5d6c2d0ee1f18294e12#diff-66df9d0cbaa60c09eb94d2865a239462

I uses nested case classes. (Outer one for non-overridable config, inner one for overridable config, inner one's apply to execute things). This will certainly be slightly unfamiliar, but it does match the problem nicely. Do you guys think that's trying to be too smart and we should rather stick rather to the style @jodersky's ensime plugin chose? I would like to standardize on one style across all plugins. The disadvantage of the current ensime plugins style is mainly that every plugin will basically have to name-prefix all of their top-level keys to avoid conflicts with other plugins. The advantage is it is more familiar (e.g. more how sbt does things) and doesn't use nested classes, so fewer of the language. Opinions?

jodersky:
I like the style you propose, @cvogt. It keeps things very focused and also avoids unused declarations in case someone overrides the top-level task. I.e. in the current situation, if someones overrides `ensime`, then all `ensimeXXX` config methods will most likely become useless.

I must admit that I barely looked through other plugins before starting work on this one. I'll refactor this plugin to take advantage of #511 and will try using the style you propose

jodersky:
@cvogt, I updated the plugin to support multi-project builds (i.e.builds with other BaseBuild dependencies) and consolidated some tasks as per our gitter discussion. Unfortunately, I was not able to fold all settings into a single public method. Here is my rationale for the various public methods:

1. `ensime`. This is the plugin's main method.
2. `ensimeConfig`. This does what a nested `config` case class would do in the design you propose. Now, I could create a dummy `apply#config` class with identical fields as EnsimeConfig, however that seems kind of redundant. Another approach would be to nest the whole plugin companion object in an `apply` case class. I am however not sure if such a nesting is the simplest approach, since the utilities may be useful by themselves as well.
3. `ensimeServerVersion`. I added this as a top-level method since many ensime config values depend on it. Since this plugin is made to work only with the latest ensime version, and since it is currently a -SNAPSHOT unstable version, I think it is safer to define it as an independent def rather than copy its value into multiple internal config values.

Please let me know if you have any suggestions to combine the 3 tasks, especially 1 and 2.

jodersky:
I removed the explicit ensime server version setting in the latest commit. The caveat is two-fold:
1. users will have to make sure that `ensimeServerVersion` and `ensimeServerJars` are coherent when customizing either one of them
2. at least java 1.8 is required

jodersky:
Check out [this diff](https://github.com/jodersky/cbt/compare/ensime...ensime-nested) to see a prototype with a single `ensime` method. I'm really not sure if the added complexity is worth it

stage2/plugins/Ensime.scala
```diff
@@ -0,0 +1,283 @@
...
+        case NonFatal(_) =>
+          logger.resolver(
+            s"ensime: could not find $classifierName of ${dependency.mavenDependency.serialize}")
+          Seq.empty
+      }
+    }
+  }
+
+  /** Find and convert all (transitive) dependencies of a build to a sequence of ENSIME projects. */
+  def findProjects(root: BaseBuild)(implicit logger: Logger, cache: JMap[AnyRef, AnyRef], classLoaderCache: ClassLoaderCache): Seq[EnsimeProject] = {
```
cvogt: You might want to coordinate with @darthorimar on the behavior of this method. He is solving the same problem for IntelliJ, we might as well agree on doing it the same way. Maybe even using the same code.

stage2/plugins/Ensime.scala
```diff
@@ -0,0 +1,283 @@
...
+    javaSources = Ensime.jdkSource.toSeq,
+    javaFlags = Ensime.baseJavaFlags,
+    projects = Ensime.findProjects(this)
+  )
+
+  /** Generate an ENSIME configuration file for this build. */
+  def ensime: File = lib.writeIfChanged(
+    projectDirectory / ".ensime",
+    ensimeConfig.toSexp
+  )
```
cvogt: if you move the body of this method into a new method called `Ensime.EnsimeConfig.apply` and move the code from `ensimeConfig` here and remove `ensimeConfig`, calling `cbt ensime` will invoke that `.apply` method. This allows not having the `ensimeConfig` member and that's what the scalatex plugin and a few others are doing. Open to discuss if this less clear than having an explicit `fooConfig` next to `foo`.

cvogt: Ah I see your comment now, you've done exactly that in that commit.

> I'm really not sure if the added complexity is worth it

Yeah, me neither.

cvogt: It IS an approach that scales to having more than just one `private` config variable. Here we only have one, which is configFile. So conformity may play into this as well. Another down side however is, that `Ensime#apply#config` is a bit of an unusual and somewhat tricky type to write down and deal with for people.

I need to experiment a bit more myself and see if reducing language features / complexity in exchange for paying (maybe only slightly) more boiler plate cost is viable everywhere.

Happy to more forward with what you have in the PR for now and unify things further at some point in the future. Having everything in the same repo will make this easy :)

cvogt:
happy to merge as is unless you want to do some adjustments?

jodersky:
I've been pondering different approaches on how to unify all settings in one public method but none seem clean so far. The biggest issue I have with wrapping `EnsimeConfig` in an `apply` is that it will suddenly be tied to a parent class instance, although it doesn't need it for its functionality. One way to work around that would be to duplicate a config class withing apply, that's not very DRY however

So in the mean time, yes pleas go ahead and merge! :)

PR #540, opened 2017-06-28
Build Info improvements #3

darthorimar:


stage2/plugins/ExportBuildInformation.scala
```diff
@@ -33,20 +34,30 @@ object BuildInformation {
...
 
   case class ScalaCompiler( version: String, jars: Seq[File] )
 
+  case class LibraryJar( jar: File, jarType: JarType.JarType )
+
+  object JarType extends Enumeration {
+    type JarType = Value
+    val Binary = Value("binary")
+    val Source = Value("source")
+  }
```
cvogt: Enums are rarely used by Scala developers. Seems like sealed traits with case objects are a bit more idiomatic.

Issue #502, opened 2017-04-25
review reproducible builds feature


cvogt:
https://contributors.scala-lang.org/t/asking-for-your-feedback-on-sbt-scala-center-announcement/738/72

PR #543, opened 2017-07-03
Make packages reproducible

jodersky:
Don't include time information in jars in order to make them
reproducible.

This allows builds to be verified with hash sums.

Also see https://reproducible-builds.org/


cvogt:
Nice!! A comment in the code liking to that URL may be nice. Will do it later when I am at a computer or if you want to be quicker ;).

jodersky:
Will do! Since this is something that can easily be forgotten, I was also thinking of adding a test to CI to make sure that packages are reproducible.

cvogt:
Sure. Nice!

jodersky:
Added a comment about the website and a test

examples/package-example/build/build.scala
```diff
@@ -0,0 +1,42 @@
...
+    for(b <- bytes) {
+      builder.append(f"$b%02x")
+    }
+    builder.toString
+  }
+
+  override def run = {
+    val pass1 = super.`package`.map(f => hash(f.toPath))
+    lib.clean(cleanFiles, true, false, false, false)
+    transientCache.clear()
```
jodersky: This cache invalidation is required here, or else the second pass will fail due to missing files. Should this be moved to the clean helper method in lib?

Also, I'm not sure if I have the same concept of what a `clean` operation should do, but IMO there should not be the need to prompt the user for a confirmation since we're only deleting generated files.

jodersky:
Linking to #502 as it is related

examples/package-example/build/build.scala
cvogt: I think it would probably be safer if we don't try to clear caches from within a single cbt run.

What if in test.scala we call `cbt package`, then read the info into memory. Call `cbt clean`, then `cbt package`, check that the file modified date is different but the jar meta data is still the same.

`clean` requiring confirmation is less about not trusting the user having called the right command, more about not trusting cbt as a fast moving code base to not silently break behavior here at some point. Once we get to the point where we have solid, in-memory filesystem test for this, we can remove the check.

jvican:
Have you tested this does not affect incremental compilation @jodersky ?

jodersky:
Sorry for not addressing your feedback earlier, I'll try to get some time this weekend.

Regarding the incremental compilation, I haven't tested this, but I don't think that the compiler would change timestamps within packaged jars. Again, this is just speculation, but seems to me that it would be a very strange approach to actually rely on filesystem write access for incremental compilation (apart from a dedicated state-store or cache, of course)

jvican:
@jodersky To the best of my understanding, the JDK does indeed cache jars using that timestamp information. What I don't actually know is which information it uses. It might be the metadata of the whole jar, or of every jar entry.

cvogt:
For what operations? Do you have a link?

jvican:
Classloading in general. Not saying it is an issue. It may be. https://github.com/scala/bug/issues/10295

jodersky:
Hmm, I'm not quite sure how incremental compilation works here. Does it mean that to check if a source needs recompilation, it checks the last modification date of its dependencies and compares it with a local cache? In case it doesn't use any checksum it should be fairly easy to spoof; I'll give it a shot.

The changes proposed here do similar things as this maven plugin https://zlika.github.io/reproducible-build-maven-plugin/ (and also what happens

cvogt:
Jorge, I think that's just scalac. I think the JDK just loads classes once if not loaded, no timestamp involved. But could be wrong.

jodersky:
That's my understanding too, at least from briefly going over the linked issue.

In that case, how big of deal would this be? Usually, local dependencies use classfiles directly rather than the ones packaged in a jar (this is the `export` setting in sbt)

examples/package-example/build/build.scala
jodersky: @cvogt, should the checksum be generated and checked in the master test file then?

jvican:
It doesn't look like something related to Scalac only:
> My analysis of the internal caching in java.util.File was incorrect. Even prior to Java 9, the last modified timestamp is used as part of the cache key.
  
This is the JDK caching jars that are classloaded.
  
In any case, this could be an intricate use case that may not affect this particular change. We also wanted to experiment with this: https://github.com/sbt/io/pull/58. So please let me know if you find any issue in Zinc. I'll fix it.

jvican:
> Does it mean that to check if a source needs recompilation, it checks the last modification date of its dependencies and compares it with a local cache?
  
This doesn't happen for jars. Jars are sha1'd and then compared. See https://github.com/sbt/zinc/blob/d4d29e8ffeecfa7c6a8e834b4ee54c4bfda9af63/zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala#L184-L186. But please, do double-check in case I'm wrong.

jodersky:
thanks for looking into this @jvican. I assume we can move forward with this change then?

PR #548, opened 2017-07-18
Giter8 plugin

darthorimar:


stage2/plugins.scala
```diff
@@ -7,6 +7,7 @@ class plugins( context: Context, scalaVersion: String ) {
     ),
     None
   )
+  final lazy val gitter = plugin( "giter8" )
```
cvogt: doesn't look like the plugin exists, no?

PR #550, opened 2017-07-19
Running specific main class

darthorimar:


stage2/BasicBuild.scala
```diff
@@ -216,6 +216,16 @@ trait BaseBuild extends BuildInterface with DependencyImplementation with SbtDep
 
   def run: ExitCode = runMain( context.args )
 
+  def runMain: ExitCode = {
+    context.args.headOption match {
+      case Some(className) => 
+        runMain(className, context.args.drop(1))
+      case None =>
+        System.err.println("No classname is specified")
```
cvogt: easier to understand would be `usage: cbt runMain <class name>`

PR #554, opened 2017-08-09
Fix gui summoning browser in headless mode

jasonholloway:
I'm using the Windows Subsystem for Linux in my adventures with CBT - and it runs headless, which the GUI tool (up till now) finds surprising and chokes on. This very humble small commit fixes that.


cvogt:
LGTM

PR #556, opened 2017-08-26
Create a ScalaXB Plugin

Jacoby6000:
Most of this seems to work, aside from a class loader issue that's preventing me from testing it.

Any suggestions for improvements welcome while I try to sort it out!

### DIFF
--- a/plugins/scalaxb/ScalaXB.scala
+++ b/plugins/scalaxb/ScalaXB.scala
@@ -1,0 +1,138 @@
+package cbt
+
+import java.io.File
+import scalaxb.{compiler => sc}
+import scalaxb.compiler.{Config => ScConfig}
+import sc.ConfigEntry._
+import cbt.ScalaXB._
+
+trait ScalaXB extends BaseBuild {
+  private val targetDirectory = projectDirectory / "target"
+
+  def generateScalaXBSources(outputDirectory: File = (targetDirectory / "src_managed"), verbose: Boolean = false) =
+    ScalaXB.GenerateScalaXBSources(sources, projectDirectory, outputDirectory, verbose)
+}
+
+object ScalaXB {
+  case class GenerateScalaXBSources(sources: Seq[File], projectRoot: File, outputDirectory: File, verbose: Boolean = false) {
+    def defaultConfig: Config =
+      Config(
+        xsdSourceDir = projectRoot / "src" / "main" / "xsd",
+        wsdlSourceDir = projectRoot / "src" / "main" / "wsdl",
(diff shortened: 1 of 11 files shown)


cvogt:
@Jacoby6000 in order to sign the CLA you need to add the email address used in your commits to your github account.

plugins/scalaxb/ScalaXB.scala
```diff
@@ -0,0 +1,138 @@
...
+  sealed trait HttpClient
+  case class Dispatch(version: String, generateAs: Boolean = false) extends HttpClient
+  case class Gigahorse(version: String, backend: Gigahorse.Backend) extends HttpClient
+  case object Async extends HttpClient
+  object Gigahorse {
+    sealed trait Backend
+    case object OkHttp extends Backend
+    case object AHC extends Backend
+  }
+}
```
cvogt: can we take the Gigahorse stuff out of the cbt package?

cvogt:
will look into the crash now :)

cvogt:
Getting closer. This seems be be caused by cbt itself using an older Scala version (2.11.8) than one of the dependencies of the plugin (2.11.11). We could temporarily fix this problem but upgrading cbt's scala version but it would also apply again whenever a new Scala version comes out. Will have to think of a solution for this. Nice find in any case. This would have popped up soon anyways.

cvogt:
btw. the way I found out is by printing the classloader in the top line of cbt in the StackTrace of the crash. `Lib.scala:70`

cvogt:
While I am still working on a more permanent fix, if you rebase your PR on this one: https://github.com/cvogt/cbt/pull/560 you should be unblocked. Upgrading scala to 2.11.11 solved the problem for now. Note that there is still a crash, but it looks like a problem in your plugin this time.

cvogt:
(#560 currently fails the build, but don't worry about that. It looks like it only affect cross-cbt-version integration, which shouldn't block you here and I'll look into as well)

plugins/scalaxb/ScalaXB.scala
Jacoby6000: I can, but it's built in to scalaxb, so the binaries will still be floating around in there.. scalaxb has a tight coupling to it's backends from what I understand.

cvogt: I mean, can we move inside of a scalaxb plugin specific object or package? Just so it doesn't lie around in cbt's global namespace

Jacoby6000:
I'll work on the tests, since we've addressed that classloader issue

PR #563, opened 2017-10-30
Export build info fixes

darthorimar:


cvogt:
cc @darthorimar fixed the tests for this

darthorimar:
I'll fix that later :)

darthorimar:
It is better to remove PrettyPrinter. There was some errors with it but I don't remember exactly with what the problem was

PR #576, opened 2017-12-25
Scalafmt - .scalafmt.conf lookup

mfolnovic:
Currently, `scalafmt` only searches for `.scalafmt.conf` in `projectDirectory`. This doesn't work well for multi-module projects where there's only one `.scalafmt.conf` in root directory.

This PR changes search logic so that it first searches for .scalafmt.conf in all parent directories (not just `projectDirectory`), then fallbacks to home directory, and then fallbacks to default config.

Code by @cvogt on Gitter: https://gitter.im/cvogt/cbt?at=5a41021f0163b0281078c1a5

### DIFF
--- a/plugins/scalafmt/Scalafmt.scala
+++ b/plugins/scalafmt/Scalafmt.scala
@@ -10,26 +10,36 @@ import java.nio.file._
 
 /** This plugin provides scalafmt support for cbt. */
 trait Scalafmt extends BaseBuild {
+  private def userHome = Option( System.getProperty("user.home") ).map(Paths.get(_))
+
   /** Reformat scala source code according to `scalafmtConfig` rules */
   def scalafmt = {
+    val configFile =
+      Stream
+        .iterate( projectDirectory )( _.getParentFile )
+        .map( Option( _ ) )
+        .takeWhile( _.nonEmpty )
+        .flatten
+        .map( _.toPath )
+        .flatMap( Scalafmt.loadConfig( _ ) )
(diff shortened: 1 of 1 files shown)


plugins/scalafmt/Scalafmt.scala
```diff
@@ -10,26 +10,36 @@ import java.nio.file._
...
+      Stream
+        .iterate( projectDirectory )( _.getParentFile )
+        .map( Option( _ ) )
+        .takeWhile( _.nonEmpty )
+        .flatten
+        .map( _.toPath )
+        .flatMap( Scalafmt.loadConfig( _ ) )
+        .headOption
+
+    val fallback = userHome flatMap ( Scalafmt.loadConfig( _ ) )
```
cvogt: Not sure this is a good idea. This would mean that the scalafmt file lives outside the project which would be problematic for any multi user git project. Do you actually use this? Otherwise how do you feel about removing this?

cvogt: Ah. I see. You just refactored an existing behavior. I like the simplification.

cvogt: Ah. I see. You just refactored an existing behavior. I like the simplification.

cvogt: Ah. I see. You just refactored an existing behavior. I like the simplification.

cvogt:
Love it!

PR #577, opened 2017-12-31
Windows script fix

sake92:
This PR fixes most of the issues with using CBT on Windows:
- cache strings are now *platform independent* as far as they can, by using `pip(unixPath: String)` (path-independent-path, just converts slashes to backslashes)
- batch script is now working properly and is *usable*

Some minor issues left:
- improve Nailgun support, `java.net.SocketException: Connection reset`
- time function
- nailgun log files are impossible to write to, when ng-server is up (`.. >> %nailgun_out% 2>> %nailgun_err%`)
- `eval` can't be used with single-quotes since CMD treats them specially... You can use double-quotes but then you can't use `String` literals 😄

### DIFF
--- a/nailgun_launcher/src/cbt/EarlyDependencies.java
+++ b/nailgun_launcher/src/cbt/EarlyDependencies.java
@@ -5,6 +5,9 @@ import java.nio.file.*;
 import java.net.*;
 import java.security.*;
 import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 import static cbt.Stage0Lib.*;
 import static cbt.NailgunLauncher.*;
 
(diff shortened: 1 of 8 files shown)


nailgun_launcher/src/cbt/EarlyDependencies.java
```diff
@@ -154,4 +157,13 @@ public EarlyDependencies(
...
+  
+  /** Platform independent path.<br>
+   * Replaces slashes with backslashes, e.g. something/bla becomes something\bla on Windows
+   */
+  private String pip(String pathWithSlashes) {
+	  // when replacing the path with \ Java treats it like escape character,
+	  // that's why we need Matcher.quoteReplacement
+	  return pathWithSlashes.replaceAll("/", Matcher.quoteReplacement(File.separator));
+  }
 }
```
cvogt: This file is auto-generated ( see comment in top line ). Editing it by hand is not the way to go.

sake92: Oh, I didn't see that. :/ Guess I'll have to find another way..

sake92: How about [this solution](https://github.com/cvogt/cbt/pull/577/files#diff-1b1258815a38bfcdf7559850f41a3037R168)?  
Btw, when I do `cbt tools cbtEarlyDependencies` it generates the file in `cbt/nailgun_launcher` and not in `cbt/nailgun_launcher/src/cbt`.. Is that expected?

cvogt: looks like a bug

cvogt:
great stuff :)

Issue #583, opened 2018-01-24
Incorrect finding of $CBT_HOME when cbt symlink is relative


buzden:
After cloning CBT and symlinking the `cbt` executable from those place that is in `$PATH`, I get the following when running `cbt`:
```
/usr/local/bin/cbt: line 109: ../../../home/buzden/cbt/realpath/realpath.sh: No such file or directory
(Note: nailgun not found. It makes CBT faster! Try 'brew install nailgun' or 'apt-get install nailgun'.)
Error: Could not find or load main class cbt.NailgunLauncher
```
Just to clarify the paths: `/home/buzden/cbt` is the directory where CBT-repository was cloned to, `/usr/local/bin/cbt` is a symlink in a directory that is in `$PATH`.

It is important that symlink was made relative, i.e. I used `ln -sf ../../../home/buzden/cbt/cbt .` command. If I had used absolute symlink, everything run okay. But every time you make an absolute symlink, one kitten dies. Let's save the kittens!

cvogt:
very nice report and PR, great stuff :)

PR #584, opened 2018-01-24
Support of relative symlinks to the cbt executable has been implemented.

buzden:
This is a fix for #583

### DIFF
--- a/cbt
+++ b/cbt
@@ -93,7 +93,7 @@ fi
 
 CWD=$(pwd)
 
-CBT_SCRIPT="$(readlink "$0")"
+CBT_SCRIPT="$(readlink -m "$0")"
 if [ "$CBT_SCRIPT" = "" ]; then
 	CBT_SCRIPT="$0"
 fi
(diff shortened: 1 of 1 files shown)


cvogt:
this broke OSX, pushing an adjustment

buzden:
Okay. But I hope you understand that '-m' also makes everything work in case when there is a symlink to a symlink to the executable (and as many levels as you want). OSX workaround does not work with it. Are there any more kinds of canonicalization except '-m' which still work for OSX?

cvogt:
I don't know. If there is a generally better solution I am all for it. For now I just made sure this doesn't break anything. You commit fixes linux, so now master is strictly better than it was before :). That's enough for me for now. As long as things improve, PRs don't need to deliver the perfect solution for what I am concerned.

buzden:
> so now master is strictly better than it was before

Exactly so, no doubt! I'm so sorry not thinking about anyone except linuxers ;-)

I'm just saying that there probably will be a problem in the future and it will raise purely on OSX.

Issue #585, opened 2018-01-26
Mains of not only objects are considered to be entrypoints


buzden:
Consider I have (accidentially or not) some `class X` which has `def main(args: Array[String]): Unit`. CBT considers this as one of the variants of possible entry points during `cbt run` unless the `class` is `abstract`. 

This behaviour is *not fatal* but a little bit *annoying* when you accidentially have written `class` instead of `object` and everything works fine with `cbt` but does not work with other tools (e.g. running with pure `scala` runner). In the code it's written that it's done because of dottydoc.

The really strange behaviour of `cbt` is when you have a <sub><sup>really strange</sup></sub> class having both constructor parameters and `def main`, something like this:
```scala
class X(whatever: Int) {
  def main(args: Array[String]): Unit = ???
}
```

Having this leads to (guess what):
```java
Exception in thread "main" java.lang.InstantiationException: X
        at java.lang.Class.newInstance(Class.java:427)
        at cbt.reflect.Module$class.typeStaticMethod(reflect.scala:168)
        ...
Caused by: java.lang.NoSuchMethodException: X.<init>()
        ...
```

So, generally, cbt's reflection sublibrary assumes that it works with classes having a constructor with empty parameters list when talking about static methods (`typeStaticMethod(Method)` function).

cvogt:
fix in #587, let's keep this open to still add a test as described there. Probably by copying cbt/test/simple, adding the suggested classes and calling something like `.mainMethods` on the build and check it doesn't throw

PR #587, opened 2018-02-12
Classes with non-trivial constructors were made to be not considered as possible main classes

buzden:
Possible fix for #585. Classes with non-public or parameterized contructors were made to be not searched by the main class searcher. This leads to
- not failing when project contains such class (with non-public or parameterized constructor and ``def main(args: Array[String])`` at the same time);
- ignoring such classes when searching a main-class, in particular, not showing such classes when there are alternatives.

And, oh, I changed the signature of the ``typeStaticMethod``, but I hope I made it more correct (since, as far as I can see, you use -``OrFail`` suffix for those methods you are expecting to be easily failing).

### DIFF
--- a/libraries/reflect/reflect.scala
+++ b/libraries/reflect/reflect.scala
@@ -164,12 +164,12 @@ trait Module {
   def typeStaticMethod[Arg, Result]( method: Method ): Option[StaticMethod[Arg, Result]] = {
     val m = method
     val instanceOption =
-      if ( m.isStatic ) Some(null)
+      if ( m.isStatic ) Some( null )
       else m.declaringClass.getConstructors.find( _.getParameterCount == 0 ).map( _.newInstance() ) // Dottydoc needs this. It's main method is not static.
     instanceOption.map( instance => StaticMethod(
       arg => m.invoke( instance, arg.asInstanceOf[AnyRef] ).asInstanceOf[Result],
       m
-    ))
+    ) )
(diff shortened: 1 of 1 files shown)


buzden:
Also, I suggest to add something like the following as a test.

```scala
object A { // Nice main class.
  def main(args: Array[String]): Unit = println("Hello, normal object!")
}

class X private { // Should not be considered as a main class.
  def main(args: Array[String]): Unit = println("Hello, private parameterless constructor!")
}

class Y(z: Int) { // Should not be considered as a main class.
  def main(args: Array[String]): Unit = println("Hello, parameterized constructor!")
}

class Z { // Okay as a main class by cbt.
  def main(args: Array[String]): Unit = println("Hello, public parameterless constructor!")
}
```

It is expected that (having this, for instance, in a single and only file of a project) ``cbt run`` does not fail and suggests you to run either ``A`` or ``Z`` as a main-class.

cvogt:
Could you give an example of what this is fixing? Maybe including the code and call that trigger it, the problem you are seeing and what you are expecting to see instead?

buzden:
Classes ``X`` and ``Y`` from above are examples of the situation when their existence in a project leads to simple failing of the cbt.

cvogt:
ah, sorry. forgot to read the related ticket.

cvogt:
ah, great. LGTM.

